PR #11, opened 2022-04-06
- adds support for query parameter attribute

baywet:
related https://github.com/microsoft/kiota/issues/1445

### DIFF
--- a/Microsoft.Kiota.Abstractions.Tests/QueryParametersBaseTests.cs
+++ b/Microsoft.Kiota.Abstractions.Tests/QueryParametersBaseTests.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Kiota.Abstractions.Tests
             var requestInfo = new RequestInformation
             {
                 HttpMethod = Method.GET,
-                UrlTemplate = "http://localhost/me?select={select}"
+                UrlTemplate = "http://localhost/me?select={%24select}"
             };
             Action<GetQueryParameters> q = x => x.Select = new[] { "id", "displayName" };
             var qParams = new GetQueryParameters();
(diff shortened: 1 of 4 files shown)


src/QueryParameterAttribute.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Microsoft.Kiota.Abstractions;
+
+/// <summary>
+/// This attribute allows mapping between the query parameter name in the template and the property name in the class.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Field, Inherited = false, AllowMultiple = false)]
+public sealed class QueryParameterAttribute : Attribute
+{
+    readonly string templateName;
```
andrueastman, line 14:
```suggestion
    private readonly string templateName;
```

Microsoft.Kiota.Abstractions.Tests/QueryParametersBaseTests.cs
```diff
@@ -13,7 +13,7 @@ public void SetsSelectQueryParameters()
             var requestInfo = new RequestInformation
             {
                 HttpMethod = Method.GET,
-                UrlTemplate = "http://localhost/me?select={select}"
+                UrlTemplate = "http://localhost/me?select={%24select}"
```
andrueastman, line 16:
```suggestion
                UrlTemplate = "http://localhost/me?%24select={select}"
```

andrueastman (commented):
Just a couple of comments from me. 
Otherwise we just need to bump version number and release notes in csproj.

Microsoft.Kiota.Abstractions.Tests/QueryParametersBaseTests.cs
baywet:
```suggestion
                UrlTemplate = "http://localhost/me{?%24select}"
```
actually this syntax makes the parameter mandatory, which is not what we want.

andrueastman (approved):

PR #13, opened 2022-04-07
- fixes a bug where empty query parameters would be passed

baywet:
I have stopped the release and caught this minor bug in my integration testing.


andrueastman:
Looks good. 
Just realized we also missed the version bump in #11 so the release won't go through.

PR #15, opened 2022-04-13
- aligns request options with configuration revamp

baywet:
depends on https://github.com/microsoft/kiota/pull/1520

## TODO

- [x] bump version
- [x] add changelog entry


src/RequestInformation.cs
```diff
@@ -75,10 +75,42 @@ public class RequestInformation
...
         public IDictionary<string, string> Headers { get; set; } = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
         /// <summary>
+        /// Adds request headers to the request.
+        /// </summary>
+        /// <param name="source">The request headers to add.</param>
+        public void AddHeaders(IDictionary<string, string> source)
+        {
+            if(source == null) return;
+            foreach(var header in source)
+                Headers.Add(header.Key, header.Value);
```
andrueastman, line 111: Do you think we should maybe use `TryAdd`? or do we want the user to possibly experience an exception?
```suggestion
                Headers.TryAdd(header.Key, header.Value);
```

src/RequestInformation.cs
```diff
@@ -75,10 +75,42 @@ public class RequestInformation
...
+                                            x => (
+                                                Name: x.GetCustomAttributes(false)
+                                                    .OfType<QueryParameterAttribute>()
+                                                    .FirstOrDefault()?.TemplateName ?? x.Name.ToFirstCharacterLowerCase(),
+                                                Value: x.GetValue(source)
+                                            )
+                                        )
+                                        .Where(x => x.Value != null && !QueryParameters.ContainsKey(x.Name)))
+            {
+                QueryParameters.Add(property.Name, property.Value);
```
andrueastman, line 96:
```suggestion
                QueryParameters.TryAdd(property.Name, property.Value);
```

src/RequestInformation.cs
baywet: added an extension method. thanks for the callout.

src/extensions/IDictionaryExtensions.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Kiota.Abstractions.Extensions
     /// <summary>
     /// Extension methods for the <see cref="IDictionary"/>
     /// </summary>
-    public static class IDictionaryExtensions
+    internal static class IDictionaryExtensions
```
andrueastman, line 14: I believe this may break the depending packages as some of the other kiota packages do reference the `TryAdd` method after the downgrade to netstandard2.0

baywet:
```suggestion
    public static class IDictionaryExtensions
```

andrueastman (approved):

andrueastman:
Todo 
- [ ] Merge after https://github.com/microsoft/kiota/pull/1520

andrueastman (approved):

PR #30, opened 2022-08-11
Default DateTime format in URL path/query to ISO 8601

andrueastman:
This PR closes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1346 and closes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1443

Changes include : - 
- Resolves `DateTimeOffset` and `DateTime` objects to be placed in URL segments or query parameters to default to the ISO 8601
- Add explicit error message in the event that URI property is accessed without setting the "baseurl" path parameter (if needed)
- Adds tests to capture and validate these scenarios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+## [1.0.0-preview.10] - 2022-08-11
+
+### Changed
+
+- DateTime instances added to the url paths to default to ISO 8601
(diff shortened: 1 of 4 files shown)


baywet (approved):
This is great! I'd appreciate if you could reflect the changes in Java/Go and create issues for Python/PHP (in kiota).
If you can't/don't want to make code changes, just let me know and simply create issues.

andrueastman:
> I'd appreciate if you could reflect the changes in Java/Go and create issues for Python/PHP (in kiota).
If you can't/don't want to make code changes, just let me know and simply create issues


No worries @baywet. I'll sort this out as it's not a big change.

PR #34, opened 2022-09-19
feature/observability

baywet:
related https://github.com/microsoft/kiota/issues/618

### DIFF
--- a/src/RequestInformation.cs
+++ b/src/RequestInformation.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using Microsoft.Kiota.Abstractions.Extensions;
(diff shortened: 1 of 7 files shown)


src/RequestInformation.cs
```diff
@@ -160,9 +161,12 @@ public void RemoveRequestOptions(params IRequestOption[] options)
         /// <param name="content">The binary stream to set as a body.</param>
         public void SetStreamContent(Stream content)
         {
+            using var activity = _activitySource?.StartActivity(nameof(SetStreamContent));
+            setRequestType(content, activity);
             Content = content;
             Headers.Add(ContentTypeHeader, BinaryContentType);
         }
+        private static ActivitySource _activitySource = new(typeof(RequestInformation).Namespace);
```
andrueastman, line 169: Should we consider passing in the version parameter as well? 
```suggestion
        private static readonly ActivitySource _activitySource = new(typeof(RequestInformation).Namespace,typeof(RequestInformation).Assembly.GetName().Version.ToString());
```

baywet: I don't think we should, it'd mean people need to update the source in the exporter configuration everytime they upgrade the library, which would make for an easy pitfall. The tracer API (API used by all the other languages, equivalent to activity source) has the version as a separate parameter for that purpose.

andrueastman (approved):

PR #43, opened 2022-10-27
- fixes a bug where request bodies that are collections of single items would not serialize properly

baywet:
related https://github.com/microsoft/kiota/issues/1927


MIchaelMainer (approved):
LGTM

PR #50, opened 2022-11-11
Handle cases of collection of IBackedModels in the InMemoryBackingStore

andrueastman:
This PR involves feedback from @Ndiritu on handling cases for nested collections of type IBackedModel in the Backingstore and is related to https://github.com/microsoft/kiota-abstractions-php/pull/13

In summary changes include.
- Updating the backing store to subscribe to the nested collection's backingstores so that updating the properties are also dirty tracked by the root backingstore.
- Updating tests to capture the scenario

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,8 @@ bld/
 # Uncomment if you have tasks that create the project's static files in wwwroot
 #wwwroot/
 
+.idea
+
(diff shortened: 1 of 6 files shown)


Ndiritu (approved):

andrueastman:
Thanks @Ndiritu

PR #55, opened 2022-12-07
- adds support for multi-valued headers

baywet:
DO NOT MERGE BEFORE https://github.com/microsoft/kiota/pull/2045

related https://github.com/microsoft/kiota/issues/2032

### DIFF
--- a/src/RequestHeaders.cs
+++ b/src/RequestHeaders.cs
@@ -106,17 +106,17 @@ public class RequestHeaders : IDictionary<string,IEnumerable<string>> {
     public IEnumerator<KeyValuePair<string, IEnumerable<string>>> GetEnumerator() => new RequestHeadersEnumerator(_headers.GetEnumerator());
     IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
     private sealed class RequestHeadersEnumerator : IEnumerator<KeyValuePair<string, IEnumerable<string>>> {
-        private readonly Dictionary<string, HashSet<string>>.Enumerator _enumerator;
-        public RequestHeadersEnumerator(Dictionary<string, HashSet<string>>.Enumerator enumerator)
+        private readonly IEnumerator _enumerator;
+        public RequestHeadersEnumerator(IEnumerator enumerator)
         {
             _enumerator = enumerator;
         }
-        public KeyValuePair<string, IEnumerable<string>> Current => new(_enumerator.Current.Key, _enumerator.Current.Value);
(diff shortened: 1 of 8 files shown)


src/RequestHeaders.cs
```diff
@@ -0,0 +1,129 @@
...
+        if(_headers.TryGetValue(headerName, out var values))
+            foreach(var headerValue in headerValues)
+                values.Add(headerValue);
+        else
+            _headers.Add(headerName, new HashSet<string>(headerValues));
+    }
+    /// <inheritdoc/>
+    public ICollection<string> Keys => _headers.Keys;
+    /// <inheritdoc/>
+    public ICollection<IEnumerable<string>> Values => _headers.Values.Cast<IEnumerable<string>>().ToList();
```
andrueastman, line 32:
```suggestion
    public ICollection<IEnumerable<string>> Values => (ICollection<IEnumerable<string>>)_headers.Values.Cast<IEnumerable<string>>();;
```
To avoid the performance hit involved creating a new List every time this is called and iterating through the items to copy them over, maybe we could do a direct cast as we already know that Values already implements [`ICollection<T>`](https://github.com/microsoft/referencesource/blob/dae14279dd0672adead5de00ac8f117dcf74c184/mscorlib/system/collections/generic/dictionary.cs#L988)

andrueastman (commented):
Thanks for taking in the suggestion to leverage on `IDictionary`

src/RequestHeaders.cs
baywet: I'm not sure about this one. I can't cast the values collection directly because if covariance as far as I know. And casting after the cast method call really depends on the internals of LINQ, which might be a faster but riskier approach.

src/RequestHeaders.cs
```diff
@@ -0,0 +1,129 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Kiota.Abstractions;
+
+/// <summary>Represents a collection of request headers.</summary>
+public class RequestHeaders : IDictionary<string,IEnumerable<string>> {
+    private Dictionary<string, HashSet<string>> _headers = new Dictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
```
baywet, line 10:
```suggestion
    private readonly Dictionary<string, HashSet<string>> _headers = new Dictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
```

src/RequestHeaders.cs
andrueastman: True. Better the safer approach for now.

PR #61, opened 2023-01-05
Fixes empty query parameters

### LINKED ISSUE #60: RequestBuilder should not add QueryParameters in the final URI if they are null or empty
Sourced from https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/issues/564

When the query parameter is not set or empty, we should not add it to the Uri. 

Possilby we need a better check at https://github.com/microsoft/kiota-abstractions-dotnet/blob/89beebb57258163e44a962f8d4222488ef82febf/src/RequestInformation.cs#L53

andrueastman:
This PR closes #60

Changes include : - 

- Add checks to the  `AddQueryParameters` method to avoid adding empty collections and string to the `QueryParameters` when using request builders.

### DIFF
--- a/src/RequestInformation.cs
+++ b/src/RequestInformation.cs
@@ -110,10 +110,9 @@ namespace Microsoft.Kiota.Abstractions
                                         )
                                         .Where(x =>  x.Value != null &&
                                                     !QueryParameters.ContainsKey(x.Name) &&
-                                                    !string.IsNullOrEmpty(x.Value.ToString())))// no need to add an empty string value
+                                                    !string.IsNullOrEmpty(x.Value.ToString()) && // no need to add an empty string value
+                                                    (x.Value is not ICollection collection || collection.Count > 0))) // no need to add empty collection
             {
-                if(property.Value is ICollection collection && collection.Count == 0)
-                    continue; // no need to add an empty collection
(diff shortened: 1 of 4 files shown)


src/RequestInformation.cs
```diff
@@ -105,8 +108,12 @@ public void AddQueryParameters(object source)
                                                 Value: x.GetValue(source)
                                             )
                                         )
-                                        .Where(x => x.Value != null && !QueryParameters.ContainsKey(x.Name)))
+                                        .Where(x =>  x.Value != null &&
+                                                    !QueryParameters.ContainsKey(x.Name) &&
+                                                    !string.IsNullOrEmpty(x.Value.ToString())))// no need to add an empty string value
             {
+                if(property.Value is ICollection collection && collection.Count == 0)
```
baywet, line 115: you can move that to the where filter

```CSharp
property.Value is not ICollection collection || collection.Count > 0
```

baywet (dismissed):
Thanks for fixing this. Would you mind also checking Java/TS/Go please?

src/RequestInformation.cs
andrueastman: Thanks for this.

baywet (approved):

PR #63, opened 2023-01-13
Adds support for nullable ref types

andrueastman:
This PR is part of https://github.com/microsoft/kiota/issues/2073

Changes include: -

- Adds nullable markers to APIs that can return null. 
- Adds nullable feature in project as well as net6.0 target to allow for proper feature functionality. Also sets `TreatWarningsAsErrors` to alleviate any potential mistakes being ignored.

### DIFF
--- a/src/Microsoft.Kiota.Abstractions.csproj
+++ b/src/Microsoft.Kiota.Abstractions.csproj
@@ -6,7 +6,7 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <AssemblyTitle>Kiota Abstractions Library for dotnet</AssemblyTitle>
     <Authors>Microsoft</Authors>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net6.0;</TargetFrameworks>
     <LangVersion>latest</LangVersion>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <RepositoryUrl>https://github.com/microsoft/kiota-abstractions-dotnet</RepositoryUrl>
@@ -14,12 +14,15 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <Deterministic>true</Deterministic>
     <VersionPrefix>1.0.0</VersionPrefix>
-    <VersionSuffix>rc.3</VersionSuffix>
+    <VersionSuffix>rc.4</VersionSuffix>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <SignAssembly>false</SignAssembly>
     <DelaySign>false</DelaySign>
     <AssemblyOriginatorKeyFile>35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <LangVersion>latest</LangVersion>
+    <Nullable>enable</Nullable>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <!-- Enable this line once we go live to prevent breaking changes -->
     <!-- <PackageValidationBaselineVersion>1.0.0</PackageValidationBaselineVersion> -->
     <PackageReleaseNotes>
--- a/src/Microsoft.Kiota.Abstractions.csproj
+++ b/src/Microsoft.Kiota.Abstractions.csproj
@@ -6,7 +6,7 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <AssemblyTitle>Kiota Abstractions Library for dotnet</AssemblyTitle>
     <Authors>Microsoft</Authors>
-    <TargetFrameworks>netstandard2.0;net6.0;</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;</TargetFrameworks>
     <LangVersion>latest</LangVersion>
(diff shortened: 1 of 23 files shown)


src/RequestHeaders.cs
```diff
@@ -35,7 +35,9 @@ public class RequestHeaders : IDictionary<string,IEnumerable<string>> {
     /// <inheritdoc/>
     public bool IsReadOnly => false;
     /// <inheritdoc/>
+#pragma warning disable CS8603 // Possible null reference return. //Can't change signature of overriden method implementation
```
baywet, line 38: Maybe return Emurable.Empty<string>() instead?

src/RequestInformation.cs
```diff
@@ -130,7 +130,7 @@ public void AddQueryParameters(object source)
         /// <summary>
         /// The Request Body.
         /// </summary>
-        public Stream Content { get; set; }
+        public Stream Content { get; set; } = Stream.Null;
```
baywet, line 133: I do think we have code that checks for nullability in the implementation. This is just a callout to double check.

src/Microsoft.Kiota.Abstractions.csproj
```diff
@@ -6,20 +6,23 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <AssemblyTitle>Kiota Abstractions Library for dotnet</AssemblyTitle>
     <Authors>Microsoft</Authors>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net6.0;</TargetFrameworks>
```
baywet, line 9: would adding netstandard2.1 not be enough to get the functionality? it'd also git us a broader compatibility range.

baywet (commented):
great work. one thing I'm not clear on is what happens to all that nullability information when we target standard 2.0? shouldn't we have conditional directive here as well?

src/RequestHeaders.cs
andrueastman: As this is an indexer, I worth itould maybe prefer to maybe throw an `KeyNotFoundException` as returning an empty array would imply that the header exists but just without values. Either way we go, I think its important to highlight that this would however break the current expectations we set out earlier [here](https://github.com/microsoft/kiota-abstractions-dotnet/blob/b0cd59539aca7df88d203fd5cae6ca1b4323db66/Microsoft.Kiota.Abstractions.Tests/RequestHeadersTests.cs#L53).

src/Microsoft.Kiota.Abstractions.csproj
andrueastman: Hey @baywet,

My understanding is that the Nullable Reference Types feature is only a compile time feature and does not really create new types but annotations on the existing reference types. The consumer of the library would then opt in/out of the warning/errors by specifying the relevant tag in the `.csproj` file. This implies that we could still distribute the library with only the `netstandard2.0` target and the nullable information would still be propagated if one opted in. (I've also verified with existing libraries like Azure.Core which [enable NRT](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Azure.Core.csproj) and still distribute [netstandard2.0 target](https://www.nuget.org/packages/Azure.Core))

The real reason for the extra target is really for OUR developer experience in order to have accurate nullable errors and context so that we do not propagate inaccurate nullable information as updating the language version in the project will indeed enable the feature but is only well supported in the later target frameworks. 

With this in mind, think `net6.0` a better target as `netstandard2.1` doesn't really add anything as `netstandard2.0` covers all that `netstandard2.1` has. In the event new language features are added, I suspect that they will most likely be built on top of `net6.0`([or its future equivalent?](https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/)) rather than netstandard which also creates more options/benefits for everyone and net framework will be covered through the `netstandard2.0` target.

baywet: Thanks for detailing things out here. I was under the impression that some of the NRT feature required runtime support, hence our compatibility conundrum. But this is probably a requirement only when you're using the attributes associated with this feature (as opposed to the ? ! symbols).
As per the target framework. The lower the version, the broader the compatibility base. Azure core, [is targeting standard 2.1](https://github.com/Azure/azure-sdk-for-net/blob/124a959609df8999f6e08265af35590a111ea6ef/sdk/core/Azure.Core/src/Azure.Core.csproj#L11).
I believe we don't need anything specific from net5 or net6 targets, [this page](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0) gives a great overview of the topic. The important part being at the very bottom, by targeting net5/6, we'd be dropping net core 3.1, mono, and xamarin, which I believe we still want to support.

src/RequestHeaders.cs
baywet: I believe that the only user of the indexer will be a client app developer. As far as I can remember the generated code only leverages the Add method (first one above), and the request adapter lists all the entries.
Dictionary throwing a key not found exception in that case, let's get closer to that implementation for consistency.

src/Microsoft.Kiota.Abstractions.csproj
andrueastman: Resolved in latest updates

src/RequestHeaders.cs
andrueastman: Resolved in latest updates

PR #65, opened 2023-01-26
- relaxes nullability tolerance when merging objects for composed types

baywet:
related https://github.com/microsoft/kiota/pull/2171
@andrueastman would you mind propagating the update to serialization libraries please?


andrueastman (approved):
Will update the serialization libs as well

PR #79, opened 2023-04-17
Adds response headers to APIException

andrueastman:
Related to https://github.com/microsoft/kiota/issues/2524

### DIFF
--- a/src/ApiException.cs
+++ b/src/ApiException.cs
@@ -2,6 +2,7 @@
 //  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
 // ------------------------------------------------------------------------------
 using System;
+using System.Collections.Generic;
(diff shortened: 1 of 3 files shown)


src/ApiException.cs
```diff
@@ -23,8 +24,14 @@ public ApiException(string message) : base(message)
...
+    
     /// <summary>
     /// The HTTP response status code.
     /// </summary>
     public int ResponseStatusCode { get; set; }
+
+    /// <summary>
+    /// The HTTP response headers.
+    /// </summary>
+    public IDictionary<string, IEnumerable<string>> ResponseHeaders { get; set; } = new Dictionary<string, IEnumerable<string>>();
```
baywet, line 36: Ordinal ignore case comparer?

baywet (changes requested):

baywet (approved):

PR #93, opened 2023-06-12
Only allow one value on certain headers

edouardpoitras:
This is related to the bug I submitted on msgraph-sdk-dotnet: https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1951

### DIFF
--- a/src/RequestHeaders.cs
+++ b/src/RequestHeaders.cs
@@ -8,7 +8,11 @@ namespace Microsoft.Kiota.Abstractions;
 /// <summary>Represents a collection of request headers.</summary>
 public class RequestHeaders : IDictionary<string,IEnumerable<string>> {
     private readonly Dictionary<string, HashSet<string>> _headers = new Dictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
-    private readonly HashSet<string> _singleValueHeaders = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "Content-Type" };
+    private readonly HashSet<string> _singleValueHeaders = new HashSet<string>(StringComparer.OrdinalIgnoreCase) {
+        "Content-Type",
+        "Content-Encoding",
+        "Content-Length"
+    };
     /// <summary>
     /// Adds values to the header with the specified name.
     /// </summary>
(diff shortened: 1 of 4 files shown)


src/RequestHeaders.cs
```diff
@@ -20,7 +21,9 @@ public class RequestHeaders : IDictionary<string,IEnumerable<string>> {
             throw new ArgumentNullException(nameof(headerValues));
         if(!headerValues.Any())
             return;
-        if(_headers.TryGetValue(headerName, out var values))
+        if(_singleValueHeaders.Contains(headerName) && _headers.ContainsKey(headerName) && _headers[headerName].Any())
+            return;
```
andrueastman, line 25: I believe we want a replacement of the original header. 
See https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1951#issuecomment-1588649698

src/RequestHeaders.cs
```diff
@@ -8,6 +8,7 @@ namespace Microsoft.Kiota.Abstractions;
 /// <summary>Represents a collection of request headers.</summary>
 public class RequestHeaders : IDictionary<string,IEnumerable<string>> {
     private readonly Dictionary<string, HashSet<string>> _headers = new Dictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
+    private readonly HashSet<string> _singleValueHeaders = new HashSet<string> { "Content-Type" };
```
andrueastman, line 11:
```suggestion
    private readonly HashSet<string> _singleValueHeaders = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "Content-Type" };
```

andrueastman (commented):
Thanks for the contribution here @edouardpoitras. Really do appreciate it.  

Apart from the comments, can you also please update the following before we merge this?

- Bump the version at https://github.com/microsoft/kiota-abstractions-dotnet/blob/main/src/Microsoft.Kiota.Abstractions.csproj#L16
- Update the [Changelog ](https://github.com/microsoft/kiota-abstractions-dotnet/blob/main/CHANGELOG.md)file
-

baywet:
Thanks for the contribution!
Aside from Andrew's comments, I wanted to provide more context here Content-Type is a [singleton field](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.3)
> Although Content-Type is defined as a singleton field, it is sometimes incorrectly generated multiple times, resulting in a combined field value that appears to be a list

See [more information](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.5) about how fields should be folded.

Looking at the other metadata fields it sounds to me that the following fields should also be added to this list:
- Content-Encoding
- Content-Length
- Content-Location (only applies to responses, maybe skipped)
- Last-Modified (same)

baywet (approved):
Thanks for making the changes. @andrueastman for final review and merge.

andrueastman (approved):
Thanks again @edouardpoitras

PR #95, opened 2023-06-22
Fix race conditions in backing store

andrueastman:
Closes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1980

Changes store dictionaries to `ConcurrentDictionary` instance to alleviate race conditions from possible concurrent writes/accesses.

### DIFF
--- a/src/store/InMemoryBackingStore.cs
+++ b/src/store/InMemoryBackingStore.cs
@@ -129,7 +129,8 @@ namespace Microsoft.Kiota.Abstractions.Store
         }
 
         /// <summary>
-        /// Adds a callback to subscribe to events in the store with the given subscription id
+        /// Adds a callback to subscribe to events in the store with the given subscription id. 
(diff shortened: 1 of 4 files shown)


src/store/InMemoryBackingStore.cs
```diff
@@ -139,7 +139,7 @@ public void Subscribe(Action<string, object?, object?> callback, string subscrip
                 throw new ArgumentNullException(nameof(subscriptionId));
             if(callback == null)
                 throw new ArgumentNullException(nameof(callback));
-            subscriptions.Add(subscriptionId, callback);
+            subscriptions.TryAdd(subscriptionId, callback);
```
baywet, line 142: I'm not sure I'm a big fan of silently failing. It's going to have big repercussions on the customer's code

andrueastman: I see this now. Have updated the call to use `AddOrReplace` now so that the store is always updated.

baywet (approved):

Issue #81, opened 2023-04-28
RequestInformation.AddQueryParameters should add parameters based on the template as well as object


seruminar:
## Using
- Kiota 1.1.3
- Microsoft.Kiota.Abstractions 1.1.1

Given:
```
//Generated ApiRequestBuilder

UrlTemplate = "{+baseurl}/api{?Parameter*}"
```
and
```
// Generated ApiRequestBuilder query parameters class

public class RequestBuilderGetQueryParameters {
    public string? Parameter { get; set; }
}
```
(Note no `QueryParameterAttribute` was generated)

and user code like
```
var response = await apiClient.Api.GetAsync(request =>
    request.QueryParameters = new()
    {
        Parameter = someParameter
    });

```

Then something like this happens:
1. The generated code has a call to `requestInfo.AddQueryParameters(requestConfig.QueryParameters)`
2. `AddQueryParameters` uses `x.Name.ToFirstCharacterLowerCase()`, which makes the `Parameter` property add itself as `parameter`
3. The later getter of `URI`, which uses `parsedUrlTemplate.SetParameter(queryStringParameter.Key, GetSanitizedValue(queryStringParameter.Value))` will not set the value due to the casing mismatch
4. The full request is missing a query parameter set at a high level using the unmodified generated code

My solution was to adjust the template, but ideally I want to keep generated code unmodified. Given all that, can we please clarify:

- Is this expected, or is a `QueryParameterAttribute` missing from the generated code?
- Is there a requirement in the OpenAPI document to have a specific casing to lead to the URL template to use camelCase?
- Is there a reason to have case-sensitivity between the URL template, the query class, and the mechanism that converts the query class to the request URL?

ichbinsteffen:
@baywet Personally I'm quite happy that my change is merged, thank you for the quick review! 
However I've gone back to investigate some more, because
I asked myself, why does the UrlTemplate from the Taven library default to case sensitive url templates and I think the answer is in the spec they mention in their repository: [URI Template Spec RFC6570](http://tools.ietf.org/html/rfc6570). 
In section 2.3 it states:

> _Variable names are case-sensitive because the name might be expanded
   within a case-sensitive URI component._
   
   So after all I think removing the `.ToFirstCharacterLowerCase()` might have been the better solution.
   Do you know why that lower-casing was implemented in the first place?

baywet:
Thank you everyone for your patience on the matter.
[...]

PR #103, opened 2023-06-29
Issue #81: Fix case insensitivity bug

ichbinsteffen:
In [Issue #81](https://github.com/microsoft/kiota-abstractions-dotnet/issues/81) @seruminar described why URIs don't contain query parameters when the models they are derived from have pascal cased property names.

Here is my short take on the problem:
    - When the URL template contains query parameters with upper case first letters and the QueryParameters contains them in lowerCase (because they are added using x.Name.ToFirstCharacterLowerCase()) then the names are not matched and the resulting URI does not contain the parameters.
    - When in the URI getter the parameters are actually set using uriTemplate.SetParameter then in the dictionary the casing does not match, and therefore they are silently failing to be set.
    - The UriTemplate ctor contains the optional setting (caseInsensitiveParameterNames) to use a case insensitive dictionary. IMHO this is the best solution because then it does not matter what kind of casing patterns the API models have.

The alternative solution to this would be to change AddQueryParameters where
`x.Name.ToFirstCharacterLowerCase()` is called for each parameter. Lowercasing the first char is what causes this bug in the first place. The strings in the dictionary would naturally match the API model from which they are collected. 
What do you think would be the better fix?

Here a little demo of the problem:
Before:
grafik
After:
grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
+## [1.2.1] - 2023-07-03
+
+### Fixed
+
+- Fixed a bug that caused the uri parameters not to be applied when the Uri template had a different casing than the parameter name that was used to set it.
 
 ## [1.2.0] - 2023-06-28
 
--- a/Microsoft.Kiota.Abstractions.Tests/RequestInformationTests.cs
+++ b/Microsoft.Kiota.Abstractions.Tests/RequestInformationTests.cs
@@ -312,6 +312,22 @@ namespace Microsoft.Kiota.Abstractions.Tests
             serializationWriterMock.Verify(x => x.WriteStringValue(It.IsAny<string>(), It.IsAny<string>()), Times.Never);
             serializationWriterMock.Verify(x => x.WriteCollectionOfPrimitiveValues(It.IsAny<string>(), It.IsAny<IEnumerable<string>>()), Times.Once);
         }
+        [Fact]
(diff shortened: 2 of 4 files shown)


baywet (changes requested):
Thanks for the contribution @ichbinsteffen 

Would you please mind adding:
- an entry to the changelog (patch, date to Monday)
- a unit test
- bump the version in the csproj

ichbinsteffen:
> @ichbinsteffen please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> [...]

Currently my legal department is looking into this. They will let me know soon if I can give a license acceptance.

baywet (approved):
Thanks for making the changes. @andrueastman for final review and merge

andrueastman (approved):

baywet:
@andrueastman I just realized the CLA was not a required check on this repo main branch policy. I think this is due to the recent migration. Can you double check the other Kiota dotnet repos please?

andrueastman:
> @andrueastman I just realized the CLA was not a required check on this repo main branch policy. I think this is due to the recent migration. Can you double check the other Kiota dotnet repos please?

Thanks for pointing this @baywet. Will look into this and resolve.

ichbinsteffen:
> @andrueastman I just realized the CLA was not a required check on this repo main branch policy. I think this is due to the recent migration. Can you double check the other Kiota dotnet repos please?

Nonetheless, our legal department allowed me to sign it. So I'll sign it in a dedicated comment.

ichbinsteffen:
@microsoft-github-policy-service agree company="Init Innovation In Traffic Systems SE"

PR #106, opened 2023-07-21
feature/multi part

baywet:
related https://github.com/microsoft/kiota/issues/220

### DIFF
--- a/src/MultipartBody.cs
+++ b/src/MultipartBody.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Kiota.Abstractions;
 /// <summary>
 /// Represents a multipart body for a request or a response.
 /// </summary>
-public class MultiPartBody : IParsable
+public class MultipartBody : IParsable
 {
     //TODO: adds to the content type header e.g. Content-Type: multipart/form-data; boundary=MyPartBoundary198374
     private Lazy<string> _boundary = new Lazy<string>(() => Guid.NewGuid().ToString("N"));
@@ -23,9 +23,14 @@ public class MultiPartBody : IParsable
     /// </summary>
     public string Boundary => _boundary.Value;
     /// <summary>
-    /// Request adapter to use to locate serialization factories.
+    /// Creates a new instance of <see cref="MultipartBody"/> with the specified request adapter.
     /// </summary>
-    public IRequestAdapter? RequestAdapter { get; set; }
+    /// <param name="requestAdapter">The request adapter to use.</param>
+    public MultipartBody(IRequestAdapter requestAdapter)
+    {
+        RequestAdapter = requestAdapter ?? throw new ArgumentNullException(nameof(requestAdapter));
+    }
+    private readonly IRequestAdapter RequestAdapter;
     /// <summary>
     /// Adds or replaces a part to the multipart body.
     /// </summary>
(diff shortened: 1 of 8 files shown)


src/MultipartBody.cs
```diff
@@ -0,0 +1,158 @@
...
+                if(part.Value.Item2 is IParsable parsable)
+                {
+                    using var partWriter = RequestAdapter.SerializationWriterFactory.GetSerializationWriter(part.Value.Item1);
+                    partWriter.WriteObjectValue(string.Empty, parsable);
+                    using var partContent = partWriter.GetSerializedContent();
+                    if(partContent.CanSeek)
+                        partContent.Seek(0, SeekOrigin.Begin);
+                    using var ms = new MemoryStream();
+                    partContent.CopyTo(ms);
+                    writer.WriteByteArrayValue(string.Empty, ms.ToArray());
```
andrueastman, line 127: Since there is a good chance that the returned stream is already a memory stream, 
- https://github.com/microsoft/kiota-serialization-json-dotnet/blob/7cbe39c2f548b59789dfe6b95fc5e799c0e7417b/src/JsonSerializationWriter.cs#L24
- https://github.com/microsoft/kiota-serialization-text-dotnet/blob/2f18bd75c07b59f41af401937b715ccb83e4c02d/src/TextSerializationWriter.cs#L16
- https://github.com/microsoft/kiota-serialization-form-dotnet/blob/9009d0f9fbbe91876a09a6038243dc499c05103d/src/FormSerializationWriter.cs#L26

We could maybe avoid the copy/new stream by doing something like this?

```cs
if(partContent is MemoryStream orginalMemoryStream)
    writer.WriteByteArrayValue(string.Empty, orginalMemoryStream.ToArray());
else
{
    using var ms = new MemoryStream();
    partContent.CopyTo(ms);
    writer.WriteByteArrayValue(string.Empty, ms.ToArray());
}
```

src/MultipartBody.cs
```diff
@@ -0,0 +1,158 @@
...
+{
+    private readonly Lazy<string> _boundary = new Lazy<string>(() => Guid.NewGuid().ToString("N"));
+    /// <summary>
+    /// The boundary to use for the multipart body.
+    /// </summary>
+    public string Boundary => _boundary.Value;
+    /// <summary>
+    /// The request adapter to use for serialization.
+    /// </summary>
+    public IRequestAdapter? RequestAdapter { get; set; }
```
andrueastman, line 27: Out of curiosity, is there a reason this is an `IRequestAdapter` instance and not a `ISerializationWriterFactory` instance? That is all we need to pull writers based on content types.

src/MultipartBody.cs
```diff
@@ -0,0 +1,158 @@
...
+                {
+                    throw new InvalidOperationException($"Unsupported type {part.Value.Item2.GetType().Name} for part {part.Key}");
+                }
+            }
+            catch(InvalidOperationException) when(part.Value.Item2 is byte[] currentBinary)
+            { // binary payload
+                writer.WriteByteArrayValue(part.Key, currentBinary);
+            }
+        }
+        writer.WriteStringValue(string.Empty, string.Empty);
```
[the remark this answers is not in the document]
baywet: done

src/MultipartBody.cs
baywet: good idea, done

src/MultipartBody.cs
baywet: interesting suggestion. The thinking here was that request adapter is what holds the configuration. And what we use in the SetContentFromParsable methods. So mostly consistency and also to reduce code gen (no need to go get the property)

andrueastman (approved):

PR #129, opened 2023-09-23
fix: issue 127 - 

### LINKED ISSUE #127: AddQueryParameters, in RequestInformation, is not adding parameters beginning with a capital to the URI
I am using Microsoft Kiota to generate an API client from an open api yaml file. In this file, one of the request parameters uses Pascal case
```
- name: TenantId
  in: query
  required: true
  schema:
    type: string
```
This was working fine when the generated API client was using version 1.3.1 of Microsoft.Kiota.Abstractions. However, when upgrading to 1.3.2, the generated uri is not including the parameter at all, which results in an error when the API is called as that parameter is required.

Here is a sample C# console app (target framework net7.0, running on Windows 10), using code I extracted from the generated API client
```
using Microsoft.Kiota.Abstractions;
using Microsoft.Kiota.Abstractions.Authentication;
using Microsoft.Kiota.Http.HttpClientLibrary;

namespace Kiota;

public class Program
{
    static void Main(string[] args)
    {
        var pathParameters = new Dictionary<string, object>
        {
            { "baseurl", "http://localhost:123" }
        };
        var requestAdapter = new HttpClientRequestAdapter(new AnonymousAuthenticationProvider())
        {
            BaseUrl = "http://localhost:123"
        };
        var requestBuilder = new TestRequestBuilder(pathParameters, requestAdapter);

        var queryParms = new TestRequestBuilder.TestRequestBuilderGetQueryParameters
        {
            TenantId = Guid.NewGuid().ToString()
        };

        var requestInfo = requestBuilder.ToGetRequestInformation(x => x.QueryParameters = queryParms);
        Console.Write(requestInfo.URI.ToString());
    }
}

public class TestRequestBuilder : BaseRequestBuilder
{
    public TestRequestBuilder(Dictionary<string, object> pathParameters, IRequestAdapter requestAdapter) : base(requestAdapter, "{+baseurl}/api/v1/Test{?TenantId*}", pathParameters)
    {
    }

    public class TestRequestBuilderGetQueryParameters
    {
        public string TenantId { get; set; }
    }

    public class TestBuilderGetRequestConfiguration
    {
        public RequestHeaders Headers { get; set; }
        public IList<IRequestOption> Options { get; set; }
        public TestRequestBuilderGetQueryParameters QueryParameters { get; set; } = new TestRequestBuilderGetQueryParameters();

        public TestBuilderGetRequestConfiguration()
        {
            Options = new List<IRequestOption>();
            Headers = new RequestHeaders();
        }
    }

    public RequestInformation ToGetRequestInformation(Action<TestBuilderGetRequestConfiguration>? requestConfiguration = default)
    {
        var requestInfo = new RequestInformation
        {
            HttpMethod = Method.GET,
            UrlTemplate = UrlTemplate,
            PathParameters = PathParameters,
        };
        requestInfo.Headers.Add("Accept", "application/problem+json");
        if (requestConfiguration != null)
        {
            var requestConfig = new TestBuilderGetRequestConfiguration();
            requestConfiguration.Invoke(requestConfig);
            requestInfo.AddQueryParameters(requestConfig.QueryParameters);
            requestInfo.AddRequestOptions(requestConfig.Options);
            requestInfo.AddHeaders(requestConfig.Headers);
        }
        return requestInfo;
    }
}
```
With 1.3.1, the uri is generated as http://localhost:123/api/v1/Test?TenantId=8d20c652-1466-444f-af2a-403a03476187

With 1.3.2 the uri is generated as http://localhost:123/api/v1/Test

Note: If I change all instances of "TenantId " to "tenantId" in the above code, then it works successfully (although this is not really a satisfactory fix as we don't really want to change the original API).

timandella:
issue 127 - Do not set first character of query parameter name to lower case

Fixes #127

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Removed the code that changed the first character of the query parameter name to lower case
+
 ## [1.3.2] - 2023-09-21
 
 ### Changed
--- a/Microsoft.Kiota.Abstractions.Tests/RequestInformationTests.cs
+++ b/Microsoft.Kiota.Abstractions.Tests/RequestInformationTests.cs
@@ -111,6 +111,26 @@ namespace Microsoft.Kiota.Abstractions.Tests
             Assert.False(requestInfo.QueryParameters.ContainsKey("select"));
         }
         [Fact]
+        public void DoesNotSetQueryParametersToLowerCaseFirstCharacter()
+        {
+            // Arrange as the request builders would
+            var requestInfo = new RequestInformation
+            {
+                HttpMethod = Method.GET,
+                UrlTemplate = "http://localhost/me{?%TenantId}"
+            };
+            Action<GetQueryParameters> q = x => x.TenantId = "Tenant1";
+            var qParams = new GetQueryParameters();
+            q.Invoke(qParams);
+
+            // Act 
+            requestInfo.AddQueryParameters(qParams);
+
+            // Assert
+            Assert.Contains("TenantId", requestInfo.QueryParameters.Keys);
+            Assert.DoesNotContain("tenantId", requestInfo.QueryParameters.Keys);
+        }
+        [Fact]
         public void SetsPathParametersOfDateTimeOffsetType()
         {
             // Arrange as the request builders would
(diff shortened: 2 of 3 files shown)


timandella:

baywet (changes requested):
Thanks for the contribution, can you add an entry in the changelog please? (Unreleased, changed)

timandella:
Add entry to the changelog

baywet (approved):
Thanks for making the changes!

PR #131, opened 2023-10-03
Add a TryAdd method to Headers

andreaTP:
Preparation work for:
 https://github.com/microsoft/kiota/pull/3401

### DIFF
--- a/src/RequestHeaders.cs
+++ b/src/RequestHeaders.cs
@@ -6,7 +6,8 @@ using System.Linq;
 namespace Microsoft.Kiota.Abstractions;
 
 /// <summary>Represents a collection of request headers.</summary>
-public class RequestHeaders : IDictionary<string,IEnumerable<string>> {
+public class RequestHeaders : IDictionary<string, IEnumerable<string>>
+{
     private readonly Dictionary<string, HashSet<string>> _headers = new Dictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
     private readonly HashSet<string> _singleValueHeaders = new HashSet<string>(StringComparer.OrdinalIgnoreCase) {
         "Content-Type",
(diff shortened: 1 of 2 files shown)


src/RequestHeaders.cs
```diff
@@ -18,7 +19,38 @@ public class RequestHeaders : IDictionary<string,IEnumerable<string>> {
...
+                values.Add(headerValue);
+        else
+            _headers.Add(headerName, new HashSet<string>(headerValues));
+    }
+    /// <summary>
+    /// Adds values to the header with the specified name if it's not already present
+    /// </summary>
+    /// <param name="headerName">The name of the header to add values to.</param>
+    /// <param name="headerValue">The values to add to the header.</param>
+    public void TryAdd(string headerName, string headerValue)
```
baywet, line 43: We technically already have a try add method for dictionaries, so maybe we could rely on this implementation here? https://github.com/microsoft/kiota-abstractions-dotnet/blob/main/src/extensions/IDictionaryExtensions.cs#L25

Also I'm not sure how the resolution between that extension method and a local method is going to work. Maybe @andrueastman can share insights here

src/RequestHeaders.cs
```diff
@@ -18,7 +19,38 @@ public class RequestHeaders : IDictionary<string,IEnumerable<string>> {
...
+    {
+        if(string.IsNullOrEmpty(headerName))
+            throw new ArgumentNullException(nameof(headerName));
+        if(headerValue == null)
+            throw new ArgumentNullException(nameof(headerValue));
+        if(!_headers.ContainsKey(headerName))
+            _headers.Add(headerName, new HashSet<string> { headerValue });
+    }
+    private void AddImpl(string headerName, bool appendIfPresent, params string[] headerValues)
```
baywet, line 52: this method is not used?

baywet (changes requested):

src/RequestHeaders.cs
andrueastman: The second parameter will be different. The disctionary here has type `<string, IEnumerable<string>` so the extension will expect `IEnumerable<string>` when we are taking in a `string` here. So this will be technically just be an overload.

src/RequestHeaders.cs
andrueastman: Looks to be duplicate of the new `Add`. Maybe we should drop one.

andreaTP: leftover, sorry for the confusion.

andreaTP: removed

baywet (changes requested):
Thanks for making the changes!
Can you add a changelog entry please? (bump minor, set the date to tomorrow)
Can you update the version prefix in the csproj please?

andreaTP:
done, should be ready now 👍

PR #132, opened 2023-10-04
Use TryAdd when setting Content-Type

andreaTP:
Forgot to actually use `TryAdd` ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.3.5] - 2023-10-05
+
+### Changed
+
+- Uses headers try add when setting the content type.
+
 ## [1.3.4] - 2023-10-04
 
 ### Changed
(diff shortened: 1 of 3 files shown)


andrueastman (commented):
Any chance you can add a version bump and changelog entry as well to release?
Otherwise, looks good to me 👍🏼

andreaTP:
> Any chance you can add a version bump and changelog entry as well to release?

I can, but I thought what I have added yesterday is already enough, let me know if I should do it anyhow!

andrueastman:
I think we should just do it. I believe the interaction with the `SetStreamContent` and other methods would be incongruent with what we are trying to achieve at https://github.com/microsoft/kiota/issues/3378 if we don't release the change.

baywet (changes requested):
Yes we'll need a new release since the other one was published

andreaTP:
It's a holiday today, I'll make the changes tomorrow :+1:

baywet (approved):
@AndreaTP no worries, enjoy your time off, I've just pushed the changes

baywet (approved):

PR #139, opened 2023-10-19
feature/trimming

baywet:
dependa on https://github.com/std-uritemplate/std-uritemplate/pull/82 being published

### DIFF
--- a/src/Microsoft.Kiota.Abstractions.csproj
+++ b/src/Microsoft.Kiota.Abstractions.csproj
@@ -34,6 +34,7 @@
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <NoWarn>$(NoWarn);NU5048</NoWarn>
+    <IsTrimmable>true</IsTrimmable>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Microsoft.Kiota.Abstractions.csproj
+++ b/src/Microsoft.Kiota.Abstractions.csproj
@@ -33,7 +33,7 @@
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
     <PackageReadmeFile>README.md</PackageReadmeFile>
-    <NoWarn>$(NoWarn);NU5048</NoWarn>
+    <NoWarn>$(NoWarn);NU5048;NETSDK1138</NoWarn>
     <IsTrimmable>true</IsTrimmable>
   </PropertyGroup>
 
--- a/src/Microsoft.Kiota.Abstractions.csproj
+++ b/src/Microsoft.Kiota.Abstractions.csproj
@@ -6,7 +6,7 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <AssemblyTitle>Kiota Abstractions Library for dotnet</AssemblyTitle>
     <Authors>Microsoft</Authors>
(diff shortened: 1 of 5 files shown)


src/Microsoft.Kiota.Abstractions.csproj
```diff
@@ -6,7 +6,7 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <AssemblyTitle>Kiota Abstractions Library for dotnet</AssemblyTitle>
     <Authors>Microsoft</Authors>
-    <TargetFrameworks>netstandard2.0;netstandard2.1;</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net5.0;net6.0;net7.0;</TargetFrameworks>
```
baywet, line 9: even though net5.0 is a deprecated target, it's the first that supports the [trimming attributes](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.dynamicallyaccessedmembersattribute?view=net-5.0). And it's still [the recommended target for libs](https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting)

What I don't know is whether we really need net6.0 and 7, from my understanding if we only added net5, any app net 5 or above would use that version.

src/RequestInformation.cs
```diff
@@ -101,10 +104,14 @@ public Uri URI
         /// Vanity method to add the query parameters to the request query parameters dictionary.
         /// </summary>
         /// <param name="source">The query parameters to add.</param>
-        public void AddQueryParameters(object source)
+#if NET5_0_OR_GREATER
+        public void AddQueryParameters<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)] T>(T source)
```
baywet, line 108: this could be breaking to people who:
- target net 5 or above
- call the method directly
- have enabled trimming

But that population is small and would not have any query parameter work properly, so I think we could rule this as a fix.

src/Microsoft.Kiota.Abstractions.csproj
```diff
@@ -34,6 +34,7 @@
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <NoWarn>$(NoWarn);NU5048</NoWarn>
```
baywet, line 36: we might need to add a suppression for NETSDK1138 because of net5.0

src/Microsoft.Kiota.Abstractions.csproj
baywet: after further testing it seems that net 5 is all we need, resolving

andrueastman: I was going to ask about target framework of net5.0 vs net6.0 since net5.0 is out of support. But it looks like the important thing is to have the trimming attributes propagated to the compiler when then the app is finally being built and the `NET5_0_OR_GREATER` flag should do the trick with the biggest target area possible.

src/RequestInformation.cs
andrueastman: Just to confirm, will this break generated code that does not specify the type parameter? For example https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/5fbfa66e9b0d7f97228ab7a99dec749f90f71810/src/Microsoft.Graph/Generated/Users/UsersRequestBuilder.cs#L126C62-L126C77

baywet: no, the type parameter should be inferred from the first argument. I don't expect code generation to need any change.

andrueastman: That makes sense. Thanks for confirming.

andrueastman (approved):

PR #141, opened 2023-10-27
feature/serialization helpers

baywet:
partial https://github.com/microsoft/kiota/issues/3406

### DIFF
--- a/src/serialization/KiotaSerializer.Serialization.cs
+++ b/src/serialization/KiotaSerializer.Serialization.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Kiota.Abstractions.Serialization;
 /// <summary>
 /// Set of helper methods for serialization
 /// </summary>
-public static class SerializationHelpers
+public static partial class KiotaSerializer
 {
     /// <summary>
     /// Serializes the given object into a string based on the content type.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.6.0] - 2023-10-31
+
+### Added
+
+- Added helper methods to facilitate serialization and deserialization of models. [microsoft/kiota#3406](https://github.com/microsoft/kiota/issues/3406)
+
 ## [1.5.0] - 2023-10-19
 
 ### Added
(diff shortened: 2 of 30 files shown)


src/serialization/KiotaSerializer.Serialization.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        using var stream = SerializeAsStream(contentType, value);
+        return GetStringFromStream(stream);
+    }
+    private static string GetStringFromStream(Stream stream)
+    {
+        using var reader = new StreamReader(stream);
+        return reader.ReadToEndAsync().ConfigureAwait(false).GetAwaiter().GetResult(); // so the asp.net projects don't get an error
+    }
+    private static ISerializationWriter GetSerializationWriter(string contentType, object value)
```
andrueastman, line 75: The value parameter isn't really needed in this method. So we can remove it from here

src/serialization/KiotaSerializer.Serialization.cs
```diff
@@ -0,0 +1,81 @@
...
+        return GetStringFromStream(stream);
+    }
+    /// <summary>
+    /// Serializes the given object into a string based on the content type.
+    /// </summary>
+    /// <typeparam name="T">Type of the object to serialize</typeparam>
+    /// <param name="contentType">Content type to serialize the object to </param>
+    /// <param name="value">The object to serialize.</param>
+    /// <returns>The serialized representation as a stream.</returns>
+    public static Stream SerializeAsStream<T>(string contentType, IEnumerable<T> value) where T : IParsable
```
andrueastman, line 52: Out of curiosity, should we have the `SerializeAsStream` create and return the stream? Or should we let the caller pass the stream and we simply write to it here?(This may also allow us to simple have overloads of the `Serialize` method instead of a `SerializeAsStream` and a `SerializeAsString` method similar to )

src/serialization/KiotaSerializer.Serialization.cs
baywet: the idea was to avoid duplicating the null check all over the place

src/serialization/KiotaSerializer.Serialization.cs
baywet: This is a design flaw we'll need to address in v2. Basically the serialization writer infrastructure doesn't allow to pass a stream to write directly to. So if we accepted a stream here, the only thing we'd be doing is copying from the one returned by the serialization writer to this one.

andrueastman: Yeah. I see that. The extra copy work may not be worth it now.

andrueastman (approved):

PR #158, opened 2023-11-30
Expand arrays when dealing with parameters in RequestInformation

Nex-Code:
Fix for https://github.com/microsoft/kiota/issues/3354

Objects with an array type need to be expanded and converted to strings otherwise Std.UriTemplate attempts to cast them to strings and fails later.


baywet (changes requested):
Thank you for the contribution!
Can you:
1. bump the patch version in the csproj
2. add an entry in the changelog (new section for the version, today's date, changed section)
Thanks!

PR #174, opened 2024-01-11
GetCollectionOf* methods of IParseNode can now return null

Irame:
Made the return type of `GetCollectionOfPrimitiveValues`, `GetCollectionOfEnumValues` and `GetCollectionOfObjectValues` in `IParseNode` nullable  to allow for null when the underlying type is not an collection type.

This change is part of the solution for the problem discussed here microsoft/kiota#3976.


Irame:

PR #177, opened 2024-01-16
Create types for untyped nodes

### LINKED ISSUE #175: Create types for untyped nodes
To support untyped JSON (https://github.com/microsoft/kiota/issues/2319), create the following types:

`UntypedObject`
`UntypedArray`
`UntypedString`
`UntypedNumber`
`UntypedBoolean`
`UntypedNull`
These types need to derive from `UntypedNode`.
`UntypedObject` needs a property called `properties` that is a map of `UntypedNode`.

Extend `IParseNode` interface with `UntypedNode? GetUntypedValue()`

Extend `ISerializationWriter` interface with `void WriteUntypedValue(string? key, UntypedNode? value)`

MartinM85:
Closes #175


MartinM85:
I've made changes in kiota-serialization-json-dotnet, but need to reference kiota-abstractions-dotnet 1.7.6 before creating a pull request.

Suppose that at the end, the SDK generator will generate something like this
```
    public class WorkbookFilterCriteria : IAdditionalDataHolder, IBackedModel, IParsable {
...
        public UntypedNode? Values {
            get { return BackingStore?.Get<UntypedNode?>("values"); }
            set { BackingStore?.Set("values", value); }
        }
...
        public virtual IDictionary<string, Action<IParseNode>> GetFieldDeserializers() {
            return new Dictionary<string, Action<IParseNode>> {
...
                {"values", n => { Values = n.GetUntypedValue(); } },
            };
        }
        public virtual void Serialize(ISerializationWriter writer) {
            _ = writer ?? throw new ArgumentNullException(nameof(writer));
...
            writer.WriteUntypedValue("values", Values);
            writer.WriteAdditionalData(AdditionalData);
        }
    }
```

instead of the current:

```
  public class WorkbookFilterCriteria : IAdditionalDataHolder, IBackedModel, IParsable {
...

        public Json? Values {
            get { return BackingStore?.Get<Json?>("values"); }
            set { BackingStore?.Set("values", value); }
        }
...
        public virtual IDictionary<string, Action<IParseNode>> GetFieldDeserializers() {
            return new Dictionary<string, Action<IParseNode>> {
...
                {"values", n => { Values = n.GetObjectValue<Json>(Json.CreateFromDiscriminatorValue); } },
            };
        }
        public virtual void Serialize(ISerializationWriter writer) {
            _ = writer ?? throw new ArgumentNullException(nameof(writer));
...
            writer.WriteObjectValue<Json>("values", Values);
            writer.WriteAdditionalData(AdditionalData);
        }
    }
```

src/serialization/UntypedArray.cs
```diff
@@ -0,0 +1,22 @@
+﻿// ------------------------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+// ------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+#if NET5_0_OR_GREATER
+#endif
```
baywet, line 7:
```suggestion
```

src/serialization/IParseNode.cs
```diff
@@ -139,5 +139,10 @@ public interface IParseNode
         /// </summary>
         /// <returns>The byte array value of the node.</returns>
         byte[]? GetByteArrayValue();
+        /// <summary>
+        /// Gets the untyped value of the node
+        /// </summary>
+        /// <returns>The untyped value of the node.</returns>
+        UntypedNode? GetUntypedValue();
```
baywet, line 146: the challenge with extending the interfaces is that it's a breaking change. We can leave this for now but we should think about a different design to avoid this breaking change, otherwise the changes will have to wait the next major version

baywet (commented):
Thanks for the contribution, a couple of remarks. I think you can implement the POC with a local reference for the json serialization and see how that turns out.

src/serialization/IParseNode.cs
MartinM85: I'm able to serialize/deserialize untyped nodes without extending the interfaces.

src/serialization/UntypedNode.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class UntypedNode : IParsable
+    {
+        /// <summary>
+        /// The value assigned to untyped node.
+        /// </summary>
+        public object? Value => null;
+        /// <summary>
+        /// The deserialization information for the current model
+        /// </summary>
+        public virtual IDictionary<string, Action<IParseNode>> GetFieldDeserializers() => new Dictionary<string, Action<IParseNode>>();
```
baywet, line 22: this will return a new dictionary every time it's called.
Could we assign a ReadOnly Dictionary to a constant once, and return that constant instead to avoid unecessary allocs?

baywet (changes requested):

src/serialization/UntypedNode.cs
MartinM85: Good point. Fixed.

src/serialization/UntypedNode.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// </summary>
+        public virtual void Serialize(ISerializationWriter writer) => _ = writer ?? throw new ArgumentNullException(nameof(writer));
+        /// <summary>
+        /// Creates a new instance of the appropriate class based on discriminator value
+        /// </summary>
+        /// <param name="parseNode">The parse node to use to read the discriminator value and create the object</param>
+        public static UntypedNode CreateFromDiscriminator(IParseNode parseNode)
+        {
+            _ = parseNode ?? throw new ArgumentNullException(nameof(parseNode));
+            return _modelObject;
```
baywet, line 38: In that case we should be creating a new instance to avoid mixing things up between multiple subsequent serializations

baywet (changes requested):

src/serialization/UntypedNode.cs
MartinM85: ok

src/serialization/UntypedNumber.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// <summary>
+    /// Represents an untyped node with number value.
+    /// </summary>
+    /// <param name="value">The number value associated with the node.</param>
+    public class UntypedNumber(string value) : UntypedNode
+    {
+        /// <summary>
+        /// The number associated with untyped number node.
+        /// </summary>
+        public new string Value { get; } = value;
```
baywet, line 16: something I missed at the beginning of the review, any specific reason why this is a string and not one of the number types?

MartinM85: At the beginning, I wanted to create specific types like UntypedFloat, UntypedDouble, UntypedLong, UntypedInteger, etc. but without the context I don't know what's the correct type.
Same for UntypedString where the `Value` can represent string, Guid, DateTime or DateTimeOffset.
If it makes sense, I can split UntypedNumber to UntypedDecimal, UntypedDouble, UntypedFloat, UntypedLong and UntypedInteger and split UntypedString to UntypedString, UntypedDateTime, UntypedDateTimeOffset and UntypedGuid

MartinM85: I've split UntypedNumber to UntypedDecimal, UntypedDouble, UntypedFloat, UntypedLong and UntypedInteger.
Draft for serialization updated: https://github.com/microsoft/kiota-serialization-json-dotnet/pull/167

andrueastman (commented):
Just a comment on this side. 
Otherwise, we need a version bump to 1.8.0 as well as a changelog entry.

src/serialization/UntypedLong.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// <summary>
+    /// Represents an untyped node with long value.
+    /// </summary>
+    /// <param name="value">The long value associated with the node.</param>
+    public class UntypedLong(long value) : UntypedNode
+    {
+        /// <summary>
+        /// The value associated with untyped long node.
+        /// </summary>
+        public new long Value { get; } = value;
```
baywet, line 16: instead of hiding the base property, which is not going to work in most other languages, could we:
1. move most of what's in "UntypedNode" to an abstract generic "BaseUntypedNode" class.
2. make all classes that derive from "UntypedNode" derive from "BaseUntypedNode" and pass the generic type parameter.
3. Remove the constructors, or add default constructors (parameterless)
4. make "UntypedNode" derive from "BaseUntypedNode", and pass object as the type.

throughts?

baywet (changes requested):

src/serialization/UntypedLong.cs
MartinM85: Can we remove UntypedNode? If all Untyped classes will derive from BaseUntypedNode?
Not sure about UntypedObject which has properties (untyped nodes but with different type)
maybe
`UntypedObject : BaseUntypedNode<IDictionary<string, UntypedNode>>`

baywet: it went through my mind as well, I wasn't sure "how much we needed it". It's worth considering we might need it in other languages because of type erasure and similar concepts.

andrueastman: In my head, the need for `UntypedNode` is really for a type to generate in the code that can encompass all types(by being the base) so that it can be evaluated at runtime. 
If `UntypedNode` is no longer a base of the rest and `BaseUntypedNode` is, we would end up generating `BaseUntypedNode` and there may be no need for `UntypedNode` because for handling the "object" case, we already have `UntypedObject ` for that.

The use of generics here at the base may lead to type conversion challenges at runtime given that the generic parameter needed is unknown at compile/generation time. E.g. if the base is declared as `BaseUntypedNode<T>` a method trying to pull out `UntypedObject` and `UntypedBoolean` what would you generate as `T`? (If we generate `BaseUntypedNode<object>`, this would be like converting a `List<bool>` to a `List<object>`.

Given @baywet points, maybe we could have this property as a method intead as `GetValue()` and the return type based on the node type and keep the structure as is? That way we can use function/method semantics similar to what we do in other generation scenarios across languages.

andrueastman: Yes. I think that can work and method overriding with covariant return types(different return types in derived classes inheriting from object should still be consistently achievable in other languages).

src/serialization/UntypedNode.cs
```diff
@@ -0,0 +1,41 @@
...
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace Microsoft.Kiota.Abstractions.Serialization
+{
+    /// <summary>
+    /// Base class for untyped node.
+    /// </summary>
+    public class UntypedNode : IParsable
```
baywet, line 14: I believe the goal was to make this one generic and abstract? or did I miss something @andrueastman?

baywet: yes, my concern with having everything return object is that we, and the SDK users, will have to go through additional casting.

src/serialization/UntypedDecimal.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// </summary>
+    /// <param name="value">The decimal value associated with the node.</param>
+    public class UntypedDecimal(decimal value) : UntypedNode
+    {
+        private readonly decimal _value = value;
+        /// <summary>
+        /// Gets the value associated with untyped decimal node.
+        /// </summary>
+        /// <returns>The value associated with untyped decimal node.</returns>
+        public override object GetValue() => _value;
```
andrueastman, line 18: I initially expected a change like this. But this seems to be only supported in C# 9(NET5 and above)
```suggestion
        public override decimal GetValue() => _value;
```

What I'd suggest is to change to below to avoid the casting of the return type as done in the json library PR.
```suggestion
        public new decimal GetValue() => _value;
```

Just to loop in @baywet . 
Moving from a property to a method should work in the various languages which should support some form of return type covariance. 
- [TS ](https://stackoverflow.com/questions/62515165/method-overriding-in-typescript-with-different-return-types-without-hack)
- [Java](https://www.geeksforgeeks.org/covariant-return-types-java/)
- [PHP](https://www.geeksforgeeks.org/covariance-and-contravariance-in-php/)
- [C#](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/covariant-returns#motivation) (though we may use method hiding till we drop netstandard support)
- [Go](https://go.dev/play/p/fsK0Gi_r4eC)
- [Python](https://mypy.readthedocs.io/en/latest/generics.html#type-variables-with-upper-bounds)

src/serialization/UntypedNode.cs
andrueastman: Totally agree on that. I was suggesting moving to a method which will resolve this and provide a consistent way across the langs using covariant return values. See comment at https://github.com/microsoft/kiota-abstractions-dotnet/pull/177/files#r1472874629

src/serialization/UntypedDecimal.cs
baywet: Thanks for doing the research work here, it helps!
I'd like us to have early implementations for Go/Java/TS before we proceed with merging the dotnet ones if you don't mind?

andrueastman: Thats totally cool. Thanks for taking a look!

baywet (approved):

baywet (approved):

andrueastman (approved):

baywet:
Just as an update to anyone watching this: we're finalizing the work on the other main languages before we merge everything in. Thank you for your patience :)

PR #184, opened 2024-01-29
Fix `RequestInformation` by removing `GetType`, prefer `typeof(T)`

IEvangelist:
Related to https://github.com/microsoft/kiota/issues/4065

### DIFF
--- a/src/RequestInformation.cs
+++ b/src/RequestInformation.cs
@@ -199,7 +199,7 @@ namespace Microsoft.Kiota.Abstractions
         private static string? GetEnumName<T>(T value) where T : Enum
 #endif
         {
-            var type = value.GetType();
+            var type = typeof(T);
 
             if(Enum.GetName(type, value) is not { } name)
                 throw new ArgumentException($"Invalid Enum value {value} for enum of type {type}");


src/RequestInformation.cs
```diff
@@ -199,7 +199,7 @@ private static object ReplaceEnumValueByStringRepresentation(object source)
         private static string? GetEnumName<T>(T value) where T : Enum
 #endif
         {
-            var type = value.GetType();
+            var type = typeof(T);
 
             if(Enum.GetName(type, value) is not { } name)
```
eerhardt, line 204:
```suggestion
            if(Enum.GetName<T>(value) is not { } name)
```

Note this API is only on `net5.0`+. So you will need a `#if`, since this still supports netstandard.

This helps in a few ways. You don't have to box `value`, and the runtime is a bit faster.

eerhardt: While we are in here, you might as fix the DynamicallyAccessedMembers attribute. I think we could take one of 2 options:

1. Change `All` to be what GetMember is annotated: https://github.com/dotnet/runtime/blob/91beecdc32a893a7e2a16e4026dea7985c3aac1d/src/libraries/System.Private.CoreLib/src/System/Type.cs#L234-L240. Basically change `All` to each individual public member.
2. We know that `T` is an enum, so we really only need to call `type.GetField(name)` instead of `type.GetMember(name)`. `EnumMemberAttribute` [can only be applied to fields](https://github.com/dotnet/runtime/blob/91beecdc32a893a7e2a16e4026dea7985c3aac1d/src/libraries/System.Runtime.Serialization.Primitives/src/System/Runtime/Serialization/EnumMemberAttribute.cs#L6-L7). So we should only need to `DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)`.

IEvangelist: Addressed...

src/serialization/KiotaSerializer.Deserialization.cs
```diff
@@ -31,7 +31,7 @@ private static Stream GetStreamFromString(string source)
     {
         var stream = new MemoryStream();
         using var writer = new StreamWriter(stream, Encoding.UTF8, 1024, true);
-        writer.WriteAsync(source).GetAwaiter().GetResult(); // so the asp.net projects don't get an error
```
baywet, line 34: please revert as this new implementation will cause errors for clients used in an asp.net mvc application which doesn't allow reading network streams synchronously (default)

src/serialization/KiotaSerializer.Serialization.cs
```diff
@@ -70,7 +70,7 @@ public static partial class KiotaSerializer
     private static string GetStringFromStream(Stream stream)
     {
         using var reader = new StreamReader(stream);
-        return reader.ReadToEndAsync().ConfigureAwait(false).GetAwaiter().GetResult(); // so the asp.net projects don't get an error
```
baywet, line 73: please revert as this new implementation will cause errors for clients used in an asp.net mvc application which doesn't allow reading network streams synchronously (default)

src/RequestInformation.cs
```diff
@@ -174,27 +172,27 @@ public void AddQueryParameters<T>(T source)
...
-        private static object ReplaceEnumValueByStringRepresentation(object source)
+        private static string ReplaceEnumValueByStringRepresentation(object source)
         {
             if(source is Enum enumValue && GetEnumName(enumValue) is string enumValueName)
             {
                 return enumValueName;
             }
             
-            return source;
+            return source.ToString()!;
```
baywet, line 192: this and the return type need to be reverted as a bunch of number types (and other cases) will serialize abnormally because of it.

https://github.com/std-uritemplate/std-uritemplate/blob/3c908404bdc8f05e6f04210c3eeb04fffb31c8ac/csharp/Std.UriTemplate.cs#L472

src/Microsoft.Kiota.Abstractions.csproj
```diff
@@ -6,7 +6,7 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <AssemblyTitle>Kiota Abstractions Library for dotnet</AssemblyTitle>
     <Authors>Microsoft</Authors>
-    <TargetFrameworks>netstandard2.0;netstandard2.1;net5.0;</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net6.0;net7.0;net8.0;</TargetFrameworks>
```
baywet, line 9: I'm not sure what value net6/7/8 add? 
And removing net5 is a binary breaking change.
Please revert.

baywet (changes requested):
Thanks for the contribution. A couple of remarks

src/serialization/KiotaSerializer.Deserialization.cs
IEvangelist: From what I can tell, this isn't a network stream, it's all local to this serializer and it's based on the `source` string and based on a `MemoryStream`. What am I missing?

src/Microsoft.Kiota.Abstractions.csproj
IEvangelist: I can add back `net5.0`, but I believe it would be best to include net6/7/8. This is an opinion based on the official guidance here: https://learn.microsoft.com/dotnet/standard/library-guidance/cross-platform-targeting.

baywet: > You should explicitly target net6.0, net7.0, or net8.0 when you want to use newer .NET APIs

We're from the conditional build we have in the code, we're only targeting APIs that are available with net5 and higher. We're not targeting any API specific to net6 or higher. Hence why I don't think those additional targets are required as long at we have net5 in here.

src/Microsoft.Kiota.Abstractions.csproj
```diff
@@ -6,7 +6,7 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <AssemblyTitle>Kiota Abstractions Library for dotnet</AssemblyTitle>
     <Authors>Microsoft</Authors>
-    <TargetFrameworks>netstandard2.0;netstandard2.1;net5.0;</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net5.0;net6.0;net7.0;net8.0;</TargetFrameworks>
```
IEvangelist, line 9:
```suggestion
    <TargetFrameworks>netstandard2.0;netstandard2.1;net5.0;</TargetFrameworks>
```

src/serialization/KiotaSerializer.Deserialization.cs
baywet: maybe this is indeed only required when reading/writing from a network stream. I can't remember the specifics here.
@andrueatman, do you remember what the issue was?

src/RequestInformation.cs
baywet: I believe the toString was called in the expand array method because when we switched to std uri template, the library wasn't properly processing arrays of anything else than strings initially. But I believe this has been resolved at this point and the call to ToString there should be removed for the reasons outlined in my previous comment. @andrueastman can you confirm?
The alternative being to add the missing types in the GetSanitizedValue method

andrueastman: What @baywet mentions is correct. A different library was used before for the Uri template matching before switching to std uri template which handled things differently causing a few issues due to different support levels(e.g handling collections/Guids and some of the custom types in this lib). 
At the end of the day, I think the std library should get an object and return a correct url based on what's right in the spec(in an ideal world, the `GetSanitizedValue` should get smaller and cease to exist).

src/serialization/KiotaSerializer.Deserialization.cs
andrueastman: I can't remember clearly as well. 
I think maybe the right thing to do here is probably keep this as is for now and look into exposing async overloads for the versions of the API that take in a stream param?

src/RequestInformation.cs
IEvangelist: Ok, @andrueastman -  What change should I employ at this point, all things considered?

src/serialization/KiotaSerializer.Deserialization.cs
IEvangelist: Yeah, I agree that we should expose async overloads. And we should avoid _sync over async_ at all costs, as it can be the cause of deadlocks and is an antipattern.

src/RequestInformation.cs
```diff
@@ -174,7 +172,7 @@ public void AddQueryParameters<T>(T source)
             }
         }
 
-        private static object ExpandArray(Array collection)
+        private static string[] ExpandArray(Array collection)
         {
             var passedArray = new string[collection.Length];
```
baywet, line 177: I believe the toString at line 182 should be removed and the return type changed to object[]

baywet (changes requested):
probably the final comment on my side

src/RequestInformation.cs
IEvangelist: I just tested that change locally and it breaks the `DoesNotExpandSecondLayerArrays` test with the following error:

> Message: 
>   System.ArgumentException : Illegal class passed as substitution, found System.Int32[]
>
> Stack Trace: 
>   UriTemplate.convertNativeTypes(Object value)
>   UriTemplate.AddExpandedValue(String prefix, Object value, StringBuilder result, Int32 maxChar, Boolean replaceReserved)
>   UriTemplate.AddValue(Nullable`1 op, String token, Object value, StringBuilder result, Int32 maxChar)
>   UriTemplate.AddListValue(Nullable`1 op, String token, IList value, StringBuilder result, Int32 maxChar, Boolean composite)
>   UriTemplate.ExpandToken(Nullable`1 op, String token, Boolean composite, Int32 maxChar, Boolean firstToken, IReadOnlyDictionary`2 substitutions, StringBuilder result, Int32 col)
>   UriTemplate.ExpandImpl(String str, IReadOnlyDictionary`2 substitutions)
>   UriTemplate.Expand(String template, IReadOnlyDictionary`2 substitutions)
>   RequestInformation.get_URI() line 103
>   RequestInformationTests.DoesNotExpandSecondLayerArrays() line 659
RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)

src/RequestInformation.cs
andrueastman: Okay with reverting

src/RequestInformation.cs
andrueastman: I think this is a difference in expectation (generating a URL that is not useful vs throwing an error). It may make sense to update the Test to expect the error as second layer arrays are not supported really. Thoughts?

baywet: yep, I've just updated this to that so we can fast track a release before the generator's release

baywet (approved):
Thanks for the contribution!

andrueastman (approved):

Issue #196, opened 2024-02-20
Add net8.0 as a supported TFM


hwoodiwiss:
At the moment, the Kiota libraries are trim-annotated for the the .NET 5 tfm, _however_ I have a use-case where an application that transiently references the Kiota libraries via https://github.com/octokit/dotnet-sdk (currently testing against a local copy updated to the latest Kiota libraries) seems to instead prefer to select the netstandard2.0 package, my guess is this is because net5.0 is out of support. This causes trimmer warnings, which as of SDK version 9, are now included by `TreatWarningsAsError`.

I suspect this is also a cause for the significantly larger than expected binary size when this library is included (an additional 80MB) though that could be more down to the Octokit library.

It would be useful for this case for the Kiota libraries to be published targetting the net8.0 tfm as well, and have all of the AOT analyzers enabled in that case via `<IsAotCompatible>`.

Also of note, as of .NET SDK 9.0.100-preview.1.24101.2, I now get the warning:

```
C:\Program Files\dotnet\sdk\9.0.100-preview.1.24101.2\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(90,5): warning NETSDK1212: IsTrimmable and EnableTrimAnalyzer are not supported for the target framework. Consider multi-targeting to a supported framework to enable trimming, and set IsTrimmable only for the supported frameworks. For example: [R:\source\repos\kiota-abstractions-dotnet\src\Microsoft.Kiota.Abstractions.csproj]
C:\Program Files\dotnet\sdk\9.0.100-preview.1.24101.2\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(90,5): warning NETSDK1212: <IsTrimmable Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))">true</IsTrimmable> [R:\source\repos\kiota-abstractions-dotnet\src\Microsoft.Kiota.Abstractions.csproj]
```

On the net5.0 builds, which indicates to me that the trim analysis on the net5.0 build is non-functional at the moment as well.

andrueastman:
Thanks for raising this @hwoodiwiss,

> seems to instead prefer to select the netstandard2.0 package,

[...]

PR #200, opened 2024-02-21
Add net6.0 and net8.0 as Target Frameworks

hwoodiwiss:
Updates target framework to include net6.0 and net8.0

Relates to #196


andrueastman:
Thanks for this. @hwoodiwiss 
I just need to validate that this generates the expected artifacts on the release side and we'll be good to go from my end.

andrueastman (approved):

PR #202, opened 2024-02-25
Reduce IParseNode Enum methods DynamicallyAccessedMembers scope

### LINKED ISSUE #201: IParseNode enum methods should be marked DynamicallyAccessedMemberTypes.PublicFields
Currently `GetCollectionOfEnumValues` and `GetEnumValue` are marked `DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)`. This causes IL3050 to be emitted by ILC at the moment, as `All` indicates that `Enum.GetValues(Type)` could be called dynamically.

Given current usage in the serialization libraries, PublicFields or PublicFields | NonPublicFields should be sufficient, and will prevent this warning.

This was indicated to me in https://github.com/dotnet/runtime/issues/98864, as I had originally thought this was a false positive.

hwoodiwiss:
- Update DynamicallyAccessMembers for IParseNode Enum methods
- Update version and changelog

This will need to be manually updated across the other repos that implement IParseNode

fixes #201


src/serialization/IParseNode.cs
```diff
@@ -101,7 +101,7 @@ public interface IParseNode
         /// </summary>
         /// <returns>The collection of enum values.</returns>
 #if NET5_0_OR_GREATER
-        IEnumerable<T?> GetCollectionOfEnumValues<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)]T>() where T : struct, Enum;
+        IEnumerable<T?> GetCollectionOfEnumValues<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields | DynamicallyAccessedMemberTypes.NonPublicFields)] T>() where T : struct, Enum;
```
andrueastman, line 104: Out of curiosity, do we also really need the `DynamicallyAccessedMemberTypes.NonPublicFields` or is the `DynamicallyAccessedMemberTypes.PublicFields` enough?

andrueastman (commented):
Thanks for this @hwoodiwiss 

Would you also be able to submit PRs for the serialization libs taking a dependency on `IParseNode`? If not, I can probably do that once we merge this one in.

src/extensions/EnumExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            // We can suppress ILC warning IL2075 as dynamic code is supported, so the field has not been trimmed
+#pragma warning disable IL2075
+            var type = value.GetType();
+
+            if(type.GetField(name)?.GetCustomAttribute<EnumMemberAttribute>() is { } runtimeTypeAttribute)
+                return runtimeTypeAttribute.Value;
+#pragma warning restore IL2075
+
+            return null;
```
MichalStrehovsky, line 45: This suppression doesn't look right.  IL2075 is a trimming warning and PublishTrimmed (as opposed to PublishAot) doesn't change what `RuntimeFeature.IsDynamicCodeSupported` because we still have a JIT.

Why do we need the special path for `!RuntimeFeature.IsDynamicCodeSupported`? I think this can just unconditionally return `return typeof(T).GetField(name)?.GetCustomAttribute<EnumMemberAttribute>()?.Value;` and not think about IsDynamicCode.

src/serialization/IParseNode.cs
hwoodiwiss: Yeah, `PublicFields` should be enough

hwoodiwiss:
> Thanks for this @hwoodiwiss
> 
> Would you also be able to submit PRs for the serialization libs taking a dependency on `IParseNode`? If not, I can probably do that once we merge this one in.

I'm happy to update the Serialization libraries for this

baywet (approved):

hwoodiwiss:
I'm going to need to take a deeper look at this, looking at the usage of `GetEnumName` we're always going to pass `Enum` as the generic type, as generic type information isn't available above `ReplaceEnumValueByStringRepresentation` so we'll have to rely on Runtime type information which will make this incompatible with AOT.

I'll revert this part of the change for now.

baywet (dismissed):

baywet:
Please go ahead with the issue. Do you think we should hold to merge this PR to get resolution? or do you believe this PR is already an improvement over the current situation?

hwoodiwiss:
> Please go ahead with the issue. Do you think we should hold to merge this PR to get resolution? or do you believe this PR is already an improvement over the current situation?

Cool, I'll open something this evening.

I think this is still an improvement, it gives a better base going forward, and this was by far the largest producer of trim warnings I saw when compiling AOT, just due to how many times it's used, plus nothing the serialization libraries are doing at the moment requires more than `DynamicallyAccessedMemberTypes.PublicFields`, and shouldn't need to.

MichalStrehovsky:
Enumerating fields on enums is always trimming/AOT safe. If the T is restricted to be `Enum`, it's fine to access *public static* fields on it and suppress the warning. See https://github.com/dotnet/runtime/issues/97737.

I.e. even if this is `GetNames<T>(T o) where T : Enum { o.GetType().GetFields(BindingFlags.Static | .Public)` will work and the warning can be suppressed. This guarantees `o` is _some enum_ and that's enough.

baywet:
Alright, thanks everyone for the context! I'll hand it over to @andrueastman for a final review, merge and publish!

hwoodiwiss:
> Enumerating fields on enums is always trimming/AOT safe. If the T is restricted to be `Enum`, it's fine to access _public static_ fields on it and suppress the warning. See [dotnet/runtime#97737](https://github.com/dotnet/runtime/issues/97737).
> 
> I.e. even if this is `GetNames<T>(T o) where T : Enum { o.GetType().GetFields(BindingFlags.Static | .Public)` will work and the warning can be suppressed. This guarantees `o` is _some enum_ and that's enough.

Ahh, okay, that makes sense. I'll amend this to take that into account this evening. Thank you for the info. In which case, after this, the abstractions library should be trimming safe.

baywet (approved):

src/RequestInformation.cs
```diff
@@ -203,11 +203,13 @@ private static object ReplaceEnumValueByStringRepresentation(object source)
...
+            if (Enum.GetName(type, value) is not { } name)
                 throw new ArgumentException($"Invalid Enum value {value} for enum of type {type}");
 
-            if(type.GetField(name)?.GetCustomAttribute<EnumMemberAttribute>() is { } attribute)
+#pragma warning disable IL2075 // Enumerating Enum fields is always trimming/AOT safe - https://github.com/dotnet/runtime/issues/97737
+            if (type.GetField(name, BindingFlags.Static | BindingFlags.Public)?.GetCustomAttribute<EnumMemberAttribute>() is { } attribute)
                 return attribute.Value;
+#pragma warning restore IL2075
 
             return name.ToFirstCharacterLowerCase();
```
eerhardt, line 214: Suppressing a warning this way only suppresses the Roslyn analyzer that runs during this library's build. The warning is still emitted for an end-user's app that uses this method.

Instead it should use the `[UnconditionalSuppressMessage]` attribute. Note that since it is best to only suppress the warning on the single call, we typically make a nested method to add the attribute to.

```suggestion
            [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2075:UnrecognizedReflectionPattern",
                Justification = "Enumerating Enum fields is always trimming/AOT safe - https://github.com/dotnet/runtime/issues/97737")]
            static string? GetEnumMemberValue(Type enumType, string name) =>
                type.GetField(name, BindingFlags.Static | BindingFlags.Public)?.GetCustomAttribute<EnumMemberAttribute>() is { } attribute ? attribute.Value : null;

            return GetEnumMemberValue(type, name) ?? name.ToFirstCharacterLowerCase();
```

hwoodiwiss: Ah, TIL thanks, I'd seen `UnconditionalSuppressMessage` around but I wasn't aware of the functional difference.

src/RequestInformation.cs
```diff
@@ -203,13 +203,15 @@ private static object ReplaceEnumValueByStringRepresentation(object source)
...
             var type = value.GetType();
 
-            if(Enum.GetName(type, value) is not { } name)
+            if (Enum.GetName(type, value) is not { } name)
                 throw new ArgumentException($"Invalid Enum value {value} for enum of type {type}");
 
-            if(type.GetField(name)?.GetCustomAttribute<EnumMemberAttribute>() is { } attribute)
-                return attribute.Value;
+            [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2075:UnrecognizedReflectionPattern",
+                Justification = "Enumerating Enum fields is always trimming/AOT safe - https://github.com/dotnet/runtime/issues/97737")]
```
eerhardt, line 210: Looks like my original suggestion didn't build. Try this:

```suggestion
#if NET5_0_OR_GREATER
            [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2070:UnrecognizedReflectionPattern",
                Justification = "Enumerating Enum fields is always trimming/AOT safe - https://github.com/dotnet/runtime/issues/97737")]
#endif

```

hwoodiwiss: Good catch, just for my reference, is the change from 2075 to 2070 just based on where the type object is originating. (method return vs parameter)?

eerhardt: Correct. See

https://github.com/dotnet/runtime/blob/main/docs/tools/illink/error-codes.md#il2070-trim-analysis-this-argument-does-not-satisfy-dynamicallyaccessedmembersattribute-in-call-to-target-method-the-parameter-source-parameter-of-method-source-method-does-not-have-matching-annotations-the-source-value-must-declare-at-least-the-same-requirements-as-those-declared-on-the-target-location-it-is-assigned-to

vs

https://github.com/dotnet/runtime/blob/main/docs/tools/illink/error-codes.md#il2075-trim-analysis-this-argument-does-not-satisfy-dynamicallyaccessedmembersattribute-in-call-to-target-method-the-return-value-of-method-source-method-does-not-have-matching-annotations-the-source-value-must-declare-at-least-the-same-requirements-as-those-declared-on-the-target-location-it-is-assigned-to

andrueastman (approved):

PR #205, opened 2024-03-07
Upgrade the `Std.UriTemplate` lib to 0.54 and add a unit test.

IEvangelist:
Upgrade the `Std.UriTemplate` lib to 0.54 and add a unit test.

Related to observed behavior here: https://github.com/std-uritemplate/std-uritemplate/issues/114

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.7.12] - 2024-03-08
+
+### Changed
+
+- Upgrade `Std.UriTemplate` NuGet package to version 0.54.
+- Add a unit test for emoji in URI template parameters.
+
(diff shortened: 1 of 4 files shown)


baywet (changes requested):
it's strange that dependabot didn't pick it up for us...
Can you also bump the patch and add a changelog entry (tomorrow, changed) so we can expedite the release please?

andrueastman:
Just to confirm, dependabot updated the dependency via https://github.com/microsoft/kiota-abstractions-dotnet/pull/203. We just need to release a new update to abstractions with the new dependency.
@IEvangelist Any chance you can make sure you rebase/sync your branch as well?

IEvangelist:
Done

baywet (changes requested):
I think we lost the version bump in the csproj somehow

andrueastman (approved):

IEvangelist:
Tagging @nickfloyd, you're going to want to upgrade to whatever version this ships in for the Ocokit .NET SDK.

PR #213, opened 2024-03-26
Implement MultipartBody filename

### LINKED ISSUE #212: Multipart body is missing the ability to add a filename
Hi there,

I am trying to use `Multipart` and `AddOrReplacePart(string, string, Stream)` to upload files to my ASP.NET API.
The API is unable to bind the submitted values to an `IFormFile` field.

After some trial-and-error I found out that it is because the request generated by kiota is missing the `filename` field in the `Content-Disposition` header.

I had to "hack" it in using this:

```csharp
multipartBody.AddOrReplacePart("Screenshot\"; Filename=\"" + filename, contentType, stream);
```

Notice the `\"` to hijack the `"` from the `WriteStringValue` call in `MultipartBody`.

Do you plan to add support for adding parts with filenames/custom content-disposition values?

ricardoboss:
Fixes #212.

This adds a class to hold information about a part added to a `MultipartBody`. In addition to that, it is now possible to pass a file name to the `AddOrReplacePart` method, that will be added to the `Content-Disposition` header when serializing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `MultipartBody` now supports an optional `fileName` parameter to specify the file name of the part. (https://github.com/microsoft/kiota-abstractions-dotnet/issues/212)
+
 ## [1.8.0] - 2024-03-18
 
 ### Added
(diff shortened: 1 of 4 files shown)


ricardoboss:

andrueastman (commented):
Thanks for the contribution @ricardoboss 

Could you please bump the version number in the csproj file as well as add a changelog entry to the Changelog.md file at the repository root?

andrueastman (approved):
Thanks for the contribution @ricardoboss

PR #221, opened 2024-04-18
#219 set license expression

### LINKED ISSUE #219: Set Licence expression on package
### Summary
I wish for the nuget packages to have the licence expression property set correctly

### Details

The licence expression property should be set to the correct licence type I.e. MIT as this will enable analysis of licences in use to occur in external tools & the license type will be shown in Nuget etc.

thompson-tomo:
Set license expression on package

Closes #219

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


andrueastman:
Thanks for this @thompson-tomo. Could you also bump the version and update the version as well? Unfortunately, the other change was already released at https://www.nuget.org/packages/Microsoft.Kiota.Abstractions/1.8.2

andrueastman (approved):

PR #236, opened 2024-05-10
Removes GetAwaiter() from KiotaSerializer.GetStringFromStream method

### LINKED ISSUE #226: Avoid using GetAwaiter().GetResult() to prevent deadlocks
Getting synchronously results from async functions is quite an issue and should be avoided. Specially in a UI thread environments as described in Stephen's [blog post](https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html).
Sample code in `KiotaSerializer.GetStreamFromString` method:
```csharp
// Some clients enforce async stream processing.
writer.WriteAsync(source).GetAwaiter().GetResult();
```

I guess once async overloads are added (#223), this code could be made pure sync and when async is required, a dedicated function can be used instead. Albeit it might cause breaking changes.

MihaMarkic:
Resolves #226

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Removes GetAwaiter() call in KiotaSerializer.GetStringFromStream and makes it fully synchronous.
+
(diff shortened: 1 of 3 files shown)


baywet (changes requested):
Thanks for the contribution. Let's also bump the patch to ensure a swift release.

PR #243, opened 2024-05-25
Removes LINQ usages from product code

kasperk81:


src/ApiClientBuilder.cs
```diff
@@ -74,26 +74,38 @@ public static IParseNodeFactory EnableBackingStoreForParseNodeFactory(IParseNode
...
         {
-            foreach(var entry in registry
-                                    .ContentTypeAssociatedFactories
-                                    .Where(x => !(x.Value is BackingStoreParseNodeFactory ||
-                                                    x.Value is ParseNodeFactoryRegistry))
-                                    .ToList()) // earlier versions of .NET do not allow direct modification of dictionary values while iterating
+            var keysToUpdate = new List<string>();
+            foreach(var entry in registry.ContentTypeAssociatedFactories)
+            {
+                if(!(entry.Value is BackingStoreParseNodeFactory || entry.Value is ParseNodeFactoryRegistry))
```
andrueastman, line 83:
```suggestion
                if(entry.Value is not (BackingStoreSerializationWriterProxyFactory or SerializationWriterFactoryRegistry))
```

src/ApiClientBuilder.cs
```diff
@@ -74,26 +74,38 @@ public static IParseNodeFactory EnableBackingStoreForParseNodeFactory(IParseNode
...
         {
-            foreach(var entry in registry
-                                    .ContentTypeAssociatedFactories
-                                    .Where(x => !(x.Value is BackingStoreSerializationWriterProxyFactory ||
-                                                    x.Value is SerializationWriterFactoryRegistry))
-                                    .ToList()) // earlier versions of .NET do not allow direct modification of dictionary values while iterating
+            var keysToUpdate = new List<string>();
+            foreach(var entry in registry.ContentTypeAssociatedFactories)
+            {
+                if(!(entry.Value is BackingStoreSerializationWriterProxyFactory || entry.Value is SerializationWriterFactoryRegistry))
```
andrueastman, line 100:
```suggestion
                if(entry.Value is not (BackingStoreSerializationWriterProxyFactory or SerializationWriterFactoryRegistry))
```

src/store/InMemoryBackingStore.cs
```diff
@@ -69,28 +67,30 @@ public void Set<T>(string key, T? value)
                 store[key] = valueToAdd;
             }
             else if(value is IBackedModel backedModel)
-            {// if its the first time adding a IBackedModel property to the store, subscribe to its BackingStore and use the events to flag the property is "dirty"
```
andrueastman, line 72: Any specific reason to get rid of the comments?

src/RequestInformation.cs
```diff
@@ -157,29 +156,52 @@ public Uri URI
...
+                    if(attribute is QueryParameterAttribute qpAttribute)
+                    {
+                        queryParameterAttribute = qpAttribute;
+                        break;
+                    }
+                }
+
+                if(queryParameterAttribute is not null)
+                {
+                    var templateName = queryParameterAttribute.TemplateName ?? propertyInfo.Name;
```
andrueastman, line 177: This may not be equivalent to the previous code, in the scenario where the `queryParameterAttribute` isn't present, we should still fall back to the `propertyInfo.Name` which won't happen as we won't enter the if clause(check currently failing tests).

src/store/InMemoryBackingStore.cs
andrueastman: Any chance you can preserve the deleted comments in the rest of the file?

andrueastman (approved):

PR #246, opened 2024-05-27
fix time formatting for other cultures

kasperk81:
it's either escaping with backslash to opt-out of colon's "special meanings" or explicitly specifying second argument `CultureInfo.GetCultureInfo("en-US")` to prevent it from rendering other separators.

see https://github.com/dotnet/runtime/issues/90300 and https://learn.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings#timeSeparator

took me three hours to figure out why the heck a test fails on one machine but not on the other..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [1.9.1] - 2024-05-24
+## [1.9.3] - 2024-05-28
+
+### Changed
+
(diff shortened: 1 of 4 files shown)


andrueastman (commented):
Thanks for the contribution here @kasperk81 

Any chance you can also add a test to validate the `ToString` as well as bump the version and update the changelog?

calebkiage:
I think for the unit test, you can set the current culture just before the test and then reset it back after the test?

andrueastman (approved):
Thanks for the contribution here.

PR #249, opened 2024-05-31
Add support for DateOnly and TimeOnly through implicit operator conversion

### LINKED ISSUE #115: Generate DateOnly type for newer .net releases
For `{ type: "string", format: "date" }` models, Kiota references a `Date` struct that is shipped in its Abstractions package.

In newer .net releases, it would be nice to use the built-in `System.DateOnly` type, so that the model integrates more naturally with the rest of the code.

0xced:
Fixes #115


andrueastman (approved):
Thanks for the contribution @0xced

PR #253, opened 2024-06-12
Add ext. methods to replace LINQ in generated code

kasperk81:
part of https://github.com/microsoft/kiota/issues/4732

### DIFF
--- a/src/extensions/IEnumerableExtensions.cs
+++ b/src/extensions/IEnumerableExtensions.cs
@@ -46,11 +46,7 @@ namespace Microsoft.Kiota.Abstractions.Extensions
             if (e is ICollection<T> collection)
             {
                 // Allocate an array with the exact size
-#if NET5_0_OR_GREATER
-                result = GC.AllocateUninitializedArray<T>(collection.Count);
-#else
-                result = new T[collection.Count];
-#endif
+                result = AllocateOnHeap(collection.Count);
                 collection.CopyTo(result, 0);
                 return result;
             }
@@ -59,17 +55,21 @@ namespace Microsoft.Kiota.Abstractions.Extensions
             int count = 0;
             foreach (var item in e) count++;
 
-            // Allocate array with the counted size
-#if NET5_0_OR_GREATER
-            result = GC.AllocateUninitializedArray<T>(count);
-#else
-            result = new T[count];
-#endif
+            result = AllocateOnHeap(count);
 
             // Second pass to copy the elements
             count = 0;
             foreach (var item in e) result[count++] = item;
             return result;
+
+            static AllocateOnHeap(int count)
+            {
+#if NET5_0_OR_GREATER
+                return GC.AllocateUninitializedArray<T>(collection.Count);
+#else
+                return new T[collection.Count];
+#endif
+            }
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/extensions/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.Kiota.Abstractions.Extensions
+{
+    /// <summary>
+    /// Extension methods for <see cref="IEnumerable"/>
+    /// </summary>
+    public static class IEnumerableExtensions
+    {
+        public static List<T> AsList<T>(this IEnumerable<T> e) => new List<T>(e);
```
baywet, line 16: Should protect against the source being null

src/extensions/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public static T[] AsArray<T>(this IEnumerable<T> e)
+        {
+            if (e is ICollection<T> collection)
```
baywet, line 20: Should protect against the source being null

src/extensions/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
```
baywet, line 21: If the source is already an array, we should probably just cast and return that.

baywet (changes requested):
Thanks for starting this. A couple of comments. 
+ Version bump
+ Changelog entry
+ Unit tests

src/extensions/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <summary>
+        /// Converts an <see cref="IEnumerable{T}"/> to a <see cref="List{T}"/>.
+        /// </summary>
+        /// <typeparam name="T">The type of elements in the collection.</typeparam>
+        /// <param name="e">The enumerable to convert.</param>
+        /// <returns>A <see cref="List{T}"/> containing the elements of the enumerable.</returns>
+        /// <exception cref="ArgumentNullException">Thrown when the input enumerable is null.</exception>
+        public static List<T> AsList<T>(this IEnumerable<T> e)
+        {
+            if (e is null) throw new ArgumentNullException(nameof(e), "Input collection cannot be null.");
```
baywet, line 25: what do you think about returning null instead to preserve the behaviour of what we have today with null propagation?

src/extensions/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+            T[]? result = null;
+
+            if (e is ICollection<T> collection)
+            {
+                // Allocate an array with the exact size
+#if NET5_0_OR_GREATER
+                result = GC.AllocateUninitializedArray<T>(collection.Count);
+#else
+                result = new T[collection.Count];
+#endif
```
baywet, line 53: could we avoid the duplication by having a small method that accepts the length as a parameter?

baywet (approved):
Thanks for making the changes! @andrueastman for final review and merge

andrueastman (approved):
👍🏼

PR #254, opened 2024-06-13
Fix version update script

andrueastman:
Related to https://github.com/microsoft/kiota/pull/4826

### DIFF
--- a/scripts/ValidateProjectVersionUpdated.ps1
+++ b/scripts/ValidateProjectVersionUpdated.ps1
@@ -56,7 +56,7 @@ Catch {
     Exit 1
 }
 
-$currentPublishedVersion = [System.Management.Automation.SemanticVersion]$nugetIndex.items[$nugetIndex.items.Count - 1].upper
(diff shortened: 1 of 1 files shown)


SilasKenneth (dismissed):

scripts/ValidateProjectVersionUpdated.ps1
```diff
@@ -56,7 +56,7 @@ Catch {
     Exit 1
 }
 
-$currentPublishedVersion = [System.Management.Automation.SemanticVersion]$nugetIndex.items[0].upper
+$currentPublishedVersion = [System.Management.Automation.SemanticVersion]$nugetIndex.items[$nugetIndex.items.Count - 1].upper
```
baywet, line 59: did the ordering from the service change under us? shouldn't we do some client side sorting first to ensure versions are in the right order?

andrueastman: Ordering did not change. Items are just split up in different groups. Will update PR to reflect feedback in kiota PR in a bit.

baywet (approved):

SilasKenneth (approved):

Issue #247, opened 2024-05-28
When MultipartBody is serialized the values are generated with CRLF


MartinM85:
I've a controller

```
[HttpPost("login")]
[ProducesResponseType(typeof(TokenResponse), StatusCodes.Status200OK)]
[ProducesResponseType(typeof(ErrorResponse), StatusCodes.Status401Unauthorized)]
[ProducesResponseType(typeof(ErrorResponse), StatusCodes.Status500InternalServerError)]
[Produces(MediaTypeNames.Application.Json)]
[AllowAnonymous]
public IActionResult Login([FromForm] Login login)
{
}
```

`Login` class has properties `Username` and `Password`

Part of **swagger.json** from which a client is generated

```
"/api/auth/login": {
      "post": {
        "tags": [
          "Auth"
        ],
        "summary": "Generates the token from the request data",
        "requestBody": {
          "content": {
            "multipart/form-data": {
              "schema": {
                "type": "object",
                "properties": {
                  "username": {
                    "type": "string",
                    "description": "The user's name"
                  },
                  "password": {
                    "type": "string",
                    "description": "The user's password"
                  }
                }
              },
              "encoding": {
                "username": {
                  "style": "form"
                },
                "password": {
                  "style": "form"
                }
              }
            }
          }
        },

```

Calling the `login` method from the client

```
var client = new ApiClient();
var multipartBody = new MultipartBody();
multipartBody.AddOrReplacePart("username", "application/x-www-form-urlencoded", "test");
multipartBody.AddOrReplacePart("password", "application/x-www-form-urlencoded", "abc");
var tokenResponse = await client.Api.Auth.Login.PostAsync(multipartBody);
```

When debugging the `login` method, I see that the value in `Username` is `"test\r\n"` and in `Password` is `"abc\r\n"`. At the first sight it seems to me that CRLF are added during serialization.

andrueastman:
Thanks for raising this @MartinM85 

Any chance you can confirm that the seriliazed payload is as expected when you run the following?

```cs
var stringValue = KiotaSerializer.SerializeAsString("multipart/form-data", multipartBody);
```

MartinM85:
Hi @andrueastman, Text Visualizer in VS shows me this

```
--605a817187c144bbae08f23dda69523a
Content-Type: application/x-www-form-urlencoded
Content-Disposition: form-data; name="username"

test

--605a817187c144bbae08f23dda69523a
[...]

PR #256, opened 2024-06-26
When MultipartBody is serialized the values are generated with CRLF

MartinM85:
Fix for #247

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.9.7] - 2024-06-26
+
+### Changed
+
+- Fix MultipartBody serialization
+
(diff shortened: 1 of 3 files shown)


andrueastman:
Any chance we can clarify things out in the issue at https://github.com/microsoft/kiota-abstractions-dotnet/issues/247 first?
I believe there is a possibility that no action is needed here..

MartinM85:
@andrueastman You are right, with `text/plain` everything works fine in the latest version. Not sure why I was using `application/x-www-form-urlencoded`

andrueastman:
@MartinM85 Thanks for confirming. I think we can close this one for now.

If it makes sense, the next action steps would be looking into whether we need to make a change in the `application/x-www-form-urlencoded` serialization writer as highlighted in https://github.com/microsoft/kiota-abstractions-dotnet/issues/247

andrueastman:
Closing this one for now to follow up on #247

PR #287, opened 2024-07-09
ci/dependencies groups

baywet:
- **ci: adds test dependencies groups to reduce dependabot noise**
- **fix: fixes csproj paths in dependabot config**


baywet:
from [the documention](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directory) directory / only scans the root directory

andrueastman (approved):

PR #291, opened 2024-07-10
SendPrimitiveAsync throws InvalidOperationException for enums

### LINKED ISSUE #274: SendPrimitiveAsync throws InvalidOperationException for enums
I've an endpoint which returns enum (C#):

    return Ok(MyEnum.Value1);

The generated code for `GetAsync` looks like this:

    return await RequestAdapter.SendPrimitiveAsync<MyEnum?>(requestInfo, errorMapping, cancellationToken).ConfigureAwait(false);

But it throws InvalidOperationException with message **error handling the response, unexpected type**

I can't find any workaround for it right now

MartinM85:
Closes #274

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
   <!-- Common default project properties for ALL projects-->
   <PropertyGroup>
-    <VersionPrefix>1.9.8</VersionPrefix>
+    <VersionPrefix>1.9.9</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <!-- This is overidden in test projects by setting to true-->
     <IsTestProject>false</IsTestProject>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -10,47 +10,4 @@
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
   </PropertyGroup>
-  <!-- Common project properties for PACKAGED projects TO BE RELEASED-->
-  <PropertyGroup Condition="('$(IsTestProject)' != 'true') and ('$(IsAnalyzerProject)' != 'true')">
-    <Authors>Microsoft</Authors>
-    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
-    <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <PackageIconUrl>http://go.microsoft.com/fwlink/?LinkID=288890</PackageIconUrl>
-    <RepositoryUrl>https://github.com/microsoft/kiota-dotnet</RepositoryUrl>
-    <PackageProjectUrl>https://aka.ms/kiota/docs</PackageProjectUrl>
-    <PackageReleaseNotes>
-        https://github.com/microsoft/kiota-dotnet/releases
-    </PackageReleaseNotes>
-    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
-    <PackageLicenseExpression>MIT</PackageLicenseExpression>
-    <PackageReadmeFile>README.md</PackageReadmeFile>
-    <EmbedUntrackedSources>true</EmbedUntrackedSources>
-    <Deterministic>true</Deterministic>
-    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
-    <SignAssembly>false</SignAssembly>
-    <DelaySign>false</DelaySign>
-    <AssemblyOriginatorKeyFile>35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <NoWarn>$(NoWarn);NU5048;NETSDK1138</NoWarn>
-    <IsTrimmable Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)','net5.0'))">true</IsTrimmable>
-    <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)','net8.0'))">true</IsAotCompatible>
-    <ContinuousIntegrationBuild Condition="'$(TF_BUILD)' == 'true'">true</ContinuousIntegrationBuild>
-  </PropertyGroup>
-  <!-- Common project properties for TEST projects-->
-  <PropertyGroup Condition="'$(IsTestProject)' == 'true'">
-    <IsPackable>false</IsPackable>
-    <OutputType>Library</OutputType>
-  </PropertyGroup>
-  <!-- Include ReadMe.md for PACKAGED projects TO BE RELEASED-->
-  <ItemGroup Condition="('$(IsTestProject)' != 'true') and ('$(IsAnalyzerProject)' != 'true')">
-    <None Include="README.md">
-      <Pack>True</Pack>
-      <PackagePath>
-      </PackagePath>
-    </None>
-  </ItemGroup>
-  <!-- Include SourceLink Dependency for packages TO BE RELEASED-->
-  <ItemGroup Condition="('$(IsTestProject)' != 'true') and ('$(IsAnalyzerProject)' != 'true')">
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" PrivateAssets="All" />
-  </ItemGroup>
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -10,4 +10,8 @@
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(IsTestProject)' == 'true'">
+    <IsPackable>false</IsPackable>
+    <OutputType>Library</OutputType>
+  </PropertyGroup>
 </Project>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,14 +1,5 @@
 <Project>
-  <PropertyGroup>
-    <VersionPrefix>1.9.9</VersionPrefix>
-    <VersionSuffix></VersionSuffix>
-    <!-- This is overidden in test projects by setting to true-->
-    <IsTestProject>false</IsTestProject>
-    <!-- This is overidden in test projects by setting to true-->
-    <IsAnalyzerProject>false</IsAnalyzerProject>
-    <LangVersion>latest</LangVersion>
-    <Nullable>enable</Nullable>
-  </PropertyGroup>
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
   <!-- Common project properties for PACKAGED projects TO BE RELEASED-->
   <PropertyGroup Condition="('$(IsTestProject)' != 'true') and ('$(IsAnalyzerProject)' != 'true')">
     <Authors>Microsoft</Authors>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,0 +1,50 @@
+<Project>
+  <PropertyGroup>
+    <VersionPrefix>1.9.9</VersionPrefix>
+    <VersionSuffix></VersionSuffix>
+    <!-- This is overidden in test projects by setting to true-->
+    <IsTestProject>false</IsTestProject>
+    <!-- This is overidden in test projects by setting to true-->
+    <IsAnalyzerProject>false</IsAnalyzerProject>
+    <LangVersion>latest</LangVersion>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+  <!-- Common project properties for PACKAGED projects TO BE RELEASED-->
+  <PropertyGroup Condition="('$(IsTestProject)' != 'true') and ('$(IsAnalyzerProject)' != 'true')">
+    <Authors>Microsoft</Authors>
+    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <PackageIconUrl>http://go.microsoft.com/fwlink/?LinkID=288890</PackageIconUrl>
+    <RepositoryUrl>https://github.com/microsoft/kiota-dotnet</RepositoryUrl>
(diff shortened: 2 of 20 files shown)


tests/http/httpClient/RequestAdapterTests.cs
```diff
@@ -535,5 +536,289 @@ public async Task ThrowsApiExceptionOnMissingMapping(HttpStatusCode statusCode)
...
+                URI = new Uri("https://example.com")
+            };
+
+            var response = await adapter.SendPrimitiveAsync<TestEnumWithFlags?>(requestInfo);
+
+            Assert.Equal(TestEnumWithFlags.Value2 | TestEnumWithFlags.Value3, response);
+        }
+
+        [Fact]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Trimming", "IL2066:The generic parameter of type or method has a DynamicallyAccessedMembersAttribute, but the value used for it can not be statically analyzed.", Justification = "<Pending>")]
```
andrueastman, line 771: I really don't think we should be suppressing this. 
If the warning shows up here, it means that this will also show up in a generated code project with trimming enabled. 

Any chance you'd know why the suppression is needed despite the attribute being present?

andrueastman (commented):
Before merging 
- update changelog in repository root.
- update version at https://github.com/microsoft/kiota-dotnet/blob/0bc980f55ddd05368e17896b619a81d56e5e5fcb/Directory.Build.props#L4

tests/http/httpClient/RequestAdapterTests.cs
MartinM85: The warning 
> Type passed to generic parameter 'ModelType' of 'Microsoft.Kiota.Http.HttpClientLibrary.HttpClientRequestAdapter.SendPrimitiveAsync<ModelType>(RequestInformation, Dictionary<String, ParsableFactory<IParsable>>, CancellationToken)' can not be statically determined and may not meet 'DynamicallyAccessedMembersAttribute' requirements. 

shows up only when nullable type is used

    var response = await adapter.SendPrimitiveAsync<TestEnumWithFlags?>(requestInfo);

I'm not familiar with trimming, but removing DynamicallyAccessedMembers from SendPrimitiveAsync resolves the warning (I've added DynamicallyAccessedMembers attribute based on [this comment](https://github.com/microsoft/kiota-http-dotnet/pull/266#discussion_r1652899675))

src/abstractions/IRequestAdapter.cs
```diff
@@ -50,23 +50,15 @@ public interface IRequestAdapter
         /// <param name="errorMapping">The error factories mapping to use in case of a failed request.</param>
         /// <param name="cancellationToken">The <see cref="CancellationToken"/> to use for cancelling the requests.</param>
         /// <returns>The deserialized primitive response model.</returns>
-#if NET5_0_OR_GREATER
```
baywet, line 53: while I understand the new implementation does not require type information anymore, I'd like for us to setup a validation project with trimming to ensure that it does not produce warnings anymore.

baywet (commented):
Thanks for porting the PR here! We'll need to merge #292 before we can merge this one in. But besides that it looks good to me.

src/abstractions/IRequestAdapter.cs
MartinM85: Should the validation project be part of this PR? Is it enough to reference projects from solution without extra code?

andrueastman: I think this is a valid suggestion. Would you be in a position to add this @MartinM85?

I believe we would need to follow the guidance at https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?pivots=dotnet-8-0#show-all-warnings-with-test-app
and create 2 projects (one targeting net6.0 and another for net8.0)

MartinM85: ok, I will do it

MartinM85: Errors from validation project:
1.

> HttpClientRequestAdapter.cs(302,29): Trim analysis error IL2072: Microsoft.Kiota.Http.HttpClientLibrary.HttpClientRequestAdapter.<SendPrimitiveAsync>d__21<ModelType>.MoveNext(): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.
> PublicFields' in call to 'Microsoft.Kiota.Abstractions.Helpers.EnumHelpers.GetEnumValue(Type, String)'. The return value of method 'System.Nullable.GetUnderlyingType(Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.

2.

> HttpClientRequestAdapter.cs(302,29): Trim analysis error IL2087: Microsoft.Kiota.Http.HttpClientLibrary.HttpClientRequestAdapter.<SendPrimitiveAsync>d__21<ModelType>.MoveNext(): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.
> PublicFields' in call to 'Microsoft.Kiota.Abstractions.Helpers.EnumHelpers.GetEnumValue(Type, String)'. The generic parameter 'ModelType' of 'Microsoft.Kiota.Http.HttpClientLibrary.HttpClientRequestAdapter.<SendPrimitiveAsync>d__21<ModelType>' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.

3.

> JsonParseNode.cs(332,25): Trim analysis error IL2087: Microsoft.Kiota.Serialization.Json.JsonParseNode.GetCollectionOfPrimitiveValues>d__23<T>.MoveNext(): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicFields' in call to 'Microsoft.Kiota.Abstractions.Helpers.EnumHelpers.GetEnumValue(Type, String)'. The generic parameter 'T' of 'Microsoft.Kiota.Serialization.Json.JsonParseNode.<GetCollectionOfPrimitiveValues>d__23<T>' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.

4.

> JsonParseNode.cs(332,25): Trim analysis error IL2072: Microsoft.Kiota.Serialization.Json.JsonParseNode.<GetCollectionOfPrimitiveValues>d__23<T>.MoveNext(): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicFields' in call to 'Microsoft.Kiota.Abstractions.Helpers.EnumHelpers.GetEnumValue(Type, String)'. The return value of method 'System.Nullable.GetUnderlyingType(Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.

5.

> JsonSerializationWriter.cs(443,38): Trim analysis error IL2075: Microsoft.Kiota.Serialization.Json.JsonSerializationWriter.WriteNonParsableObjectValue<T>(String, T): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicProperties' in call to 'System.Type.GetProperties()'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.

6.

> .nuget\packages\azure.core\1.40.0\lib\net6.0\Azure.Core.dll : error IL2104: Assembly 'Azure.Core' produced trim warnings. For more information see https://aka.ms/dotnet-illink/libraries

7.

> .nuget\packages\system.memory.data\1.0.2\lib\netstandard2.0\System.Memory.Data.dll : error IL2104: Assembly 'System.Memory.Data' produced trim warnings. For more information see https://aka.ms/dotnet-illink/libraries

8.

> \.nuget\packages\microsoft.net.illink.tasks\7.0.100-1.23401.1\build\Microsoft.NET.ILLink.targets(86,5): error NETSDK1144: Optimizing assemblies for size failed. Optimization can be disabled by setting the PublishTrimmed property to false.

Errors 1-4 are related to this PR. I don't think that modifying unit tests (https://github.com/microsoft/kiota-dotnet/pull/291#discussion_r1672044115) is good, but currently I've no idea how to fix it.

Error 5 can be fixed in this PR

Error 6-8 are out of the scope of this PR

@andrueastman Should be publishing of validation projects part of Azure Pipeline (ci-build.yml)?

andrueastman: > @andrueastman Should be publishing of validation projects part of Azure Pipeline (ci-build.yml)?

I think adding a project here with `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` enabled should be enough as it will make the build fail. Any chance you can push the changes as well so that we may try out the changes?

> Errors 1-4 are related to this PR. 

I believe the issue is here that a function like `GetCollectionOfPrimitiveValues` will receive a type T, but then we will call `Nullable.GetUnderlyingType` to give us another type which is then passed `EnumHelpers.GetEnumValue`. So statically, the chain doesn't have the same annotations. 

Maybe we could try to move the calling of `Nullable.GetUnderlyingType` to be inside `EnumHelpers.GetEnumValue` so that the type passed is the same type annotated at the original calling function?

tests/trimming/Microsoft.Kiota.Trimming.Validation/Microsoft.Kiota.Trimming.Validation.csproj
```diff
@@ -0,0 +1,28 @@
...
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\src\abstractions\Microsoft.Kiota.Abstractions.csproj" />
+    <ProjectReference Include="..\..\..\src\authentication\azure\Microsoft.Kiota.Authentication.Azure.csproj" />
+    <ProjectReference Include="..\..\..\src\http\httpClient\Microsoft.Kiota.Http.HttpClientLibrary.csproj" />
+    <ProjectReference Include="..\..\..\src\serialization\form\Microsoft.Kiota.Serialization.Form.csproj" />
+    <ProjectReference Include="..\..\..\src\serialization\json\Microsoft.Kiota.Serialization.Json.csproj" />
+    <ProjectReference Include="..\..\..\src\serialization\multipart\Microsoft.Kiota.Serialization.Multipart.csproj" />
+    <ProjectReference Include="..\..\..\src\serialization\text\Microsoft.Kiota.Serialization.Text.csproj" />
+    <TrimmerRootAssembly Include="Microsoft.Kiota.Abstractions" />
```
baywet, line 20: From reading the docs I understand that those instructions should be enough to get everything analyzed and that we don't need a generated client to ensure it. But I'd like your opinion on the matter.
https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?pivots=dotnet-8-0

baywet (commented):
Thanks for adding the trimming testing project!

src/abstractions/IRequestAdapter.cs
MartinM85: Moving Nullable.GetUnderlyingType to EnumHelpers.GetEnumValue resolved trimming errors. So, now only figure out how make unit tests working.

MartinM85: @andrueastman Just a question. Any reason to reference ILLink.* in test projects?

image

It seems to me that ILLink is added automatically due to 

```
<IsTrimmable Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)','net5.0'))">true</IsTrimmable>
<IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)','net8.0'))">true</IsAotCompatible>
```

in Directory.Build.props.

If I set explicitly in test projects

```
<IsTrimmable>false</IsTrimmable>
<IsAotCompatible>false</IsAotCompatible>
```

Trimming warnings in test projects disappear.

Seems like settings from

```
 <PropertyGroup Condition="('$(IsTestProject)' != 'true') and ('$(IsAnalyzerProject)' != 'true')">
```

Is applied also to test projects

andrueastman: That's odd. What you can probably do is add a `Directory.Build.Props` in the `src` folder and mobe the entire section in this clause there. (tests projects won't be in this folder....)
```xml
 <PropertyGroup Condition="('$(IsTestProject)' != 'true') and ('$(IsAnalyzerProject)' != 'true')">
```

tests/trimming/Microsoft.Kiota.Trimming.Validation/Microsoft.Kiota.Trimming.Validation.csproj
andrueastman: > TrimmerRootAssembly ensures that every part of the library is analyzed. It tells the trimmer that this assembly is a "root". A "root" assembly means the trimmer analyzes every call in the library and traverses all code paths that originate from that assembly.

Looking at the same, and I agree with this. 
We can stick with this for now and re-assess whether we want to maintain a generated client in the test in the project if we get feedback of any gaps in this approach.

src/Directory.Build.props
```diff
@@ -0,0 +1,50 @@
+<Project>
+  <PropertyGroup>
+    <VersionPrefix>1.9.9</VersionPrefix>
```
baywet, line 3: I'm a bit confused to why the version is at two places now?

MartinM85: By default, MSBUILD doesn't merge Directory.Build.props. Fixed

Directory.Build.props
```diff
@@ -10,47 +10,4 @@
...
-    <NoWarn>$(NoWarn);NU5048;NETSDK1138</NoWarn>
-    <IsTrimmable Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)','net5.0'))">true</IsTrimmable>
-    <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)','net8.0'))">true</IsAotCompatible>
-    <ContinuousIntegrationBuild Condition="'$(TF_BUILD)' == 'true'">true</ContinuousIntegrationBuild>
-  </PropertyGroup>
-  <!-- Common project properties for TEST projects-->
-  <PropertyGroup Condition="'$(IsTestProject)' == 'true'">
-    <IsPackable>false</IsPackable>
-    <OutputType>Library</OutputType>
-  </PropertyGroup>
```
baywet, line 43: it seems to me we've lost this block in the migration?

MartinM85: fixed

baywet (approved):
LGTM. Thanks for the continuous work on this one. @andrueastman for final review and merge.

andrueastman (approved):
Look good to me. 

Really appreciate the work here @MartinM85

PR #295, opened 2024-07-12
Fixes release

andrueastman:
Follow up to #291 

Update script target to position of `Directory.Build.props` to ensure SignAssembly attributes are located correctly.


baywet (approved):

andrueastman:
We also need this change before releasing. The trimming project does not inherit some props.

Issue #290, opened 2024-07-10
[Dotnet] Rollout a Kiota.Bundle package


andrueastman:
- [ ] Add a new package/project for the Kiota.Bundle that pulls/depends in the abstractions libraries. 
- [ ] Expose a derived request adapter to auto-register the serialization pieces.

Related to https://github.com/microsoft/kiota/issues/4636

PR #297, opened 2024-07-16
Adds bundle package

andrueastman:
Closes https://github.com/microsoft/kiota-dotnet/issues/290


src/bundle/KiotaRequestAdapter.cs
```diff
@@ -0,0 +1,47 @@
...
+using Microsoft.Kiota.Abstractions;
+using Microsoft.Kiota.Abstractions.Authentication;
+using Microsoft.Kiota.Abstractions.Serialization;
+using Microsoft.Kiota.Http.HttpClientLibrary;
+using Microsoft.Kiota.Serialization.Form;
+using Microsoft.Kiota.Serialization.Json;
+using Microsoft.Kiota.Serialization.Multipart;
+using Microsoft.Kiota.Serialization.Text;
+
+namespace Microsoft.Kiota
```
andrueastman, line 15: [Open for feedback]
This is exposed in the `Microsoft.Kiota` namespace

src/bundle/KiotaRequestAdapter.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    /// <summary>
+    /// The <see cref="IRequestAdapter"/> implementation that derived from <see cref="HttpClientRequestAdapter"/> with registrations configured.
+    /// </summary>
+    public class KiotaRequestAdapter : HttpClientRequestAdapter
```
andrueastman, line 20: [Open for feedback]
Naming. Is `KiotaRequestAdapter` okat for the derived instance? Or `DefaultRequestAdapter`? cc @sebastienlevert cc @baywet

src/bundle/Microsoft.Kiota.Bundle.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <Description>Kiota Bundle package providing default implementations for client setup.</Description>
+    <AssemblyTitle>Kiota Bundle package for dotnet</AssemblyTitle>
+    <!-- NET 5 target to be removed on next major version-->
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net5.0;net6.0;net8.0</TargetFrameworks>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\abstractions\Microsoft.Kiota.Abstractions.csproj" />
+    <ProjectReference Include="..\authentication\azure\Microsoft.Kiota.Authentication.Azure.csproj" />
```
baywet, line 14:
```suggestion
```

baywet (changes requested):

src/bundle/Microsoft.Kiota.Bundle.csproj
andrueastman: Correct. No need for including the auth library.

baywet (dismissed):

src/bundle/KiotaRequestAdapter.cs
baywet: The package name and the namespace name are generally aligned. Unless we have a good reason not to follow that convention, I'd suggest we do.

src/bundle/KiotaRequestAdapter.cs
baywet: Yeah default was the only other idea I could come up with as well. I don't have a strong opinion here.

src/bundle/KiotaRequestAdapter.cs
andrueastman: This may also not work for other languages I believe. So, I'll update this in an upcoming update.

src/bundle/KiotaRequestAdapter.cs
andrueastman: Looping in @maisarissi and @sebastienlevert before resolving this one.

maisarissi: I think I like more DefaultRequestAdapter, but I would be ok with both.

andrueastman: DefaultRequestAdapter it is...

baywet (approved):

src/bundle/README.md
```diff
@@ -0,0 +1,38 @@
+# Kiota Bundle Library for dotnet
+
+The Kiota Bundle Library for dotnet is the dotnet library providing default implementations for client setup. 
+The package provides a request adapter implementation with defaults serialization libraries setup fo use with a generated Kiota client.
+
+Read more about Kiota [here](https://github.com/microsoft/kiota/blob/main/README.md).
+
+## Using the Abstractions Library
```
andrueastman, line 8:
```suggestion
## Using the BundleLibrary
```

src/bundle/README.md
```diff
@@ -0,0 +1,38 @@
+# Kiota Bundle Library for dotnet
+
+The Kiota Bundle Library for dotnet is the dotnet library providing default implementations for client setup. 
+The package provides a request adapter implementation with defaults serialization libraries setup fo use with a generated Kiota client.
+
+Read more about Kiota [here](https://github.com/microsoft/kiota/blob/main/README.md).
+
+## Using the BundleLibrary
```
andrueastman, line 8:
```suggestion
## Using the Bundle Library
```

baywet (dismissed):

baywet (dismissed):

baywet (approved):

PR #298, opened 2024-07-16
Use invariant culture for (de)serialization

MartinM85:
Unit tests discovered several issues:

1. `FormSerializationWriter.WriteAnyValue` writes array as string
Test `WriteAdditionalData_AreWrittenCorrectly()`
`{ "prop12", new byte[] { 2, 4, 6 } }` is serialized as `prop12=System.Byte%5B%5D`

2. `Time` class has two constructors
`TextParseNode.GetTimeValue` parses `DateTime` value from string and calls `Time(DateTime)` ctor. In this case the `Date` property of `DateTime` contains current date.
If you create a new instance of `Time` by calling `Time(int hour, int minute, int second)` ctor as part of unit test, then `Date` property has year, month and day set to 1. 
When a unit test is created, `Assert.Equal` will fail, because it compares also `DateTime` properties, which can be different.
I would suggest to get rid of `DateTime` property and `Time(DateTime)` from `Time`. Seems to me better than overridden Equals method.

3. Issue `JsonSerializer.Serialize` inside `JsonSerializationWriter.WriteFloatValue` and `JsonSerializationWriter.WriteDoubleValue` serialize for .net462. Unit tests fail
Double

```
Assert.Equal() Failure: Strings differ
              ↓ (pos 3)
Expected: "36.8"
Actual:   "36.799999999999997"
              ↑ (pos 3)
```
Float
```
Assert.Equal() Failure: Strings differ
              ↓ (pos 3)
Expected: "36.8"
Actual:   "36.7999992"
              ↑ (pos 3)
```
I think that net462 is used in tests projects, because xunit runner doesn't support .netstandard2.x

4. Any reason why are `GetEnumValue` and `GetCollectionOfEnumValues` defined explicitly in `FormParseNode`?


src/abstractions/Time.cs
```diff
@@ -26,6 +27,29 @@ public struct Time
...
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+#if NET6_0_OR_GREATER
+            return HashCode.Combine(Hour, Minute, Second);
+#else
+            int hash = 17;
+            hash = hash * 23 + Hour.GetHashCode();
+            hash = hash * 23 + Minute.GetHashCode();
+            hash = hash * 23 + Second.GetHashCode();
```
baywet, line 49: should the exponents here be different from one another?

src/abstractions/Time.cs
```diff
@@ -26,6 +27,29 @@ public struct Time
         /// <returns>A new <see cref="TimeOnly"/> structure whose hours, minutes, seconds and milliseconds are equal to those of the supplied time.</returns>
         public static implicit operator TimeOnly(Time time) => new(time.DateTime.Hour, time.DateTime.Minute, time.DateTime.Second, time.DateTime.Millisecond);
 #endif
+        /// <summary>
+        /// Indicates whether the current object is equal to another object of the same type.
+        /// </summary>
+        /// <param name="other">An object to compare with this object.</param>
+        /// <returns>true if the current object is equal to the other parameter; otherwise, false.</returns>
+        public bool Equals(Time other) => (Hour, Minute, Second) == (other.Hour, other.Minute, other.Second);
```
baywet, line 35: I'm not sure what this syntax does? reference comparison over tuples? or member wise comparison?

src/serialization/form/FormParseNode.cs
```diff
@@ -143,21 +144,21 @@ public IEnumerable<T> GetCollectionOfPrimitiveValues<T>()
         }
     }
     /// <inheritdoc/>
-    public DateTimeOffset? GetDateTimeOffsetValue() => DateTimeOffset.TryParse(DecodedValue, out var result) ? result : null;
+    public DateTimeOffset? GetDateTimeOffsetValue() => DateTimeOffset.TryParse(DecodedValue, CultureInfo.InvariantCulture, DateTimeStyles.None, out var result) ? result : null;
```
baywet, line 147: shouldn't the style be roundtrip here? https://learn.microsoft.com/en-us/dotnet/api/system.globalization.datetimestyles?view=net-8.0

src/serialization/form/FormParseNode.cs
```diff
@@ -143,21 +144,21 @@ public IEnumerable<T> GetCollectionOfPrimitiveValues<T>()
...
     /// <inheritdoc/>
-    public Date? GetDateValue() => DateTime.TryParse(DecodedValue, out var result) ? new Date(result) : null;
+    public Date? GetDateValue() => DateTime.TryParse(DecodedValue, CultureInfo.InvariantCulture, DateTimeStyles.None, out var result) ? new Date(result) : null;
     /// <inheritdoc/>
-    public decimal? GetDecimalValue() => decimal.TryParse(DecodedValue, out var result) ? result : null;
+    public decimal? GetDecimalValue() => decimal.TryParse(DecodedValue, NumberStyles.Number, CultureInfo.InvariantCulture, out var result) ? result : null;
```
baywet, line 151: shouldn't the style be none here?

baywet (commented):
This is great work! I left a couple of questions for you.

baywet:
> I think that net462 is used in tests projects, because xunit runner doesn't support .netstandard2.x

The tests should run fine for both netfx462 and net8

> Any reason why are GetEnumValue and GetCollectionOfEnumValues defined explicitly in FormParseNode?

No, it was probably the auto-suggestion from vscode at the time

src/serialization/form/FormParseNode.cs
MartinM85: With None TryParse returns false

src/serialization/form/FormParseNode.cs
MartinM85: yes, it makes sense

src/serialization/form/FormParseNode.cs
baywet: In specific scenarios or for all scenarios?

MartinM85: `None` will work only for numbers without decimals.

src/abstractions/Time.cs
MartinM85: Exponents can be the same

tests/serialization/json/JsonSerializationWriterTests.cs
```diff
@@ -323,5 +330,311 @@ public void WritesSampleObjectValueWithUntypedProperties()
...
+            using var jsonSerializationWriter = new JsonSerializationWriter();
+
+            // Act
+            jsonSerializationWriter.WriteDecimalValue(null, value);
+            var contentStream = jsonSerializationWriter.GetSerializedContent();
+            using var reader = new StreamReader(contentStream, Encoding.UTF8);
+            var serializedString = reader.ReadToEnd();
+
+            // Assert
+            Assert.Equal("36.8", serializedString);
```
andrueastman, line 477: Looks like runtimes before net3.0 didn't have assurances that double/float did not guarantee returning round trippable strings when formatted. https://devblogs.microsoft.com/dotnet/floating-point-parsing-and-formatting-improvements-in-net-core-3-0/

Since this is a platform specific issue that was fixed in later runtimes, I would suggest using a `#ifdef` here to have an assertion for netfx to keep the validation here.

src/abstractions/Time.cs
andrueastman: I'm okay leaving it as is unless @baywet has any concerns here.

andrueastman (commented):
@MartinM85 We'll also need a changelog entry/version bump here as well. Otherwise looks good to me.

src/abstractions/Time.cs
baywet: I just realized that you multiply the previous hash value by the exponent.
 This is not usually how I do it. I usually multiply the new value by the new exponent. 
I guess in that case the math works if you're using the same exponent.
I don't have any more reservations. Thank you for the discussion on this topic. 
Resolving.

src/serialization/json/JsonParseNode.cs
```diff
@@ -148,17 +150,11 @@ public JsonParseNode(JsonElement node, KiotaJsonSerializationContext jsonSeriali
             if(_jsonNode.ValueKind != JsonValueKind.String)
                 return null;
 
-            if(_jsonNode.TryGetDateTimeOffset(out var dateTimeOffset))
+            if(TryGetUsingTypeInfo(_jsonNode, _jsonSerializerContext.DateTimeOffset, out var dateTimeOffset))
```
baywet, line 153: We should probably implement the same for other date formats. What do you think @andrueastman ?

MartinM85:
Still one opened question:
Is it expected that `FormSerializationWriter.WriteAnyValue` writes array as string
`new byte[] { 2, 4, 6 }` is serialized as `System.Byte%5B%5D`

baywet:
> Still one opened question: Is it expected that `FormSerializationWriter.WriteAnyValue` writes array as string? `new byte[] { 2, 4, 6 }` is serialized as `System.Byte%5B%5D`

I think the challenge here is there is no agreed upon and standardized way to represent by arrays for the form of mediatype.
This is probably why we don't do anything specific here and it ends up being the dotnet representation of this type.
Would it be better to base64 encode it instead?

baywet:
This probably says it all....

> The application/x-www-form-urlencoded format is in many ways an aberrant monstrosity, the result of many years of implementation accidents and compromises leading to a set of requirements necessary for interoperability, but in no way representing good design practices.

[source](https://url.spec.whatwg.org/#application/x-www-form-urlencoded)

MartinM85:
> > Still one opened question: Is it expected that `FormSerializationWriter.WriteAnyValue` writes array as string? `new byte[] { 2, 4, 6 }` is serialized as `System.Byte%5B%5D`
> 
> I think the challenge here is there is no agreed upon and standardized way to represent by arrays for the form of mediatype. This is probably why we don't do anything specific here and it ends up being the dotnet representation of this type. Would it be better to base64 encode it instead?

I was asking because FormSerializationWriter.WriteAnyValue writes IEnumerable, but not an array

src/serialization/json/JsonParseNode.cs
andrueastman: I agree. We probably should. 
The benefit here being that if we use the `_jsonSerializerContext` for parsing first so that customization is possible by passing a custom context. (Currentl not the case for the DateTime and Time types...)

andrueastman:
> I was asking because FormSerializationWriter.WriteAnyValue writes IEnumerable, but not an array

I believe arrays implement `IEnumerable`. 

> Is it expected that FormSerializationWriter.WriteAnyValue writes array as string?

One thing to note is that `WriteAnyValue` is private and not publicly exposed. It is therefore always called from a "nested" perspective (you'll probably be serializing a property of an object) . Given that that Form Serialization doesn't really support nested objects, the method will always write primitives. 
What we could probably do is add a case for the `byte []` that will base64 encode it as @baywet suggested.

https://github.com/microsoft/kiota-dotnet/blob/e06f0868375f666f5d4e647e016e218bd4dd6714/src/serialization/form/FormSerializationWriter.cs#L189

MartinM85:
> > I was asking because FormSerializationWriter.WriteAnyValue writes IEnumerable, but not an array
> 
> I believe arrays implement `IEnumerable`.

array implements IEnumerable, but `case` is for IEnumerable<object> which won't work for arrays

src/serialization/json/JsonParseNode.cs
MartinM85: Updated for Date and Time

baywet:
Yes it was probably a matter of the type assertion engine not matching in this case and just adding a new one so it explicitly matches. 
I can see that you have already done that. Thank you for making the change

tests/serialization/json/Converters/JsonDateConverter.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace Microsoft.Kiota.Serialization.Json.Tests.Converters;
+
+/// <summary>
+/// Converts a Date object or value to/from JSON.
+/// </summary>
+public class JsonDateConverter : JsonConverter<Date>
+{
+    /// <inheritdoc />
+    public override Date Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        => reader.TokenType == JsonTokenType.Null
```
baywet, line 16: Isn't this reversed?

tests/serialization/json/Converters/JsonDateConverter.cs
```diff
@@ -0,0 +1,29 @@
...
+            ? new Date()
+            : ReadInternal(ref reader);
+
+    /// <inheritdoc />
+    public override void Write(Utf8JsonWriter writer, Date value, JsonSerializerOptions options)
+        => WriteInternal(writer, value);
+
+    private static Date ReadInternal(ref Utf8JsonReader reader)
+        => new Date(DateTime.ParseExact(reader.GetString()!, "dd---MM---yyyy", CultureInfo.InvariantCulture));
```
baywet, line 25: Could we have another condition instead of a bang operator here?

tests/serialization/json/Converters/JsonDateConverter.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    private static void WriteInternal(Utf8JsonWriter writer, Date value)
+        => writer.WriteStringValue($"{value.Day}---{value.Month}---{value.Year}");
```
baywet, line 28: I'm not sure I understand why we have extra dashes?

tests/serialization/json/Converters/JsonDateTimeOffsetConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Microsoft.Kiota.Serialization.Json.Tests.Converters;
+
+/// <summary>
+/// Converts a DateTimeOffset object or value to/from JSON.
+/// </summary>
+public class JsonDateTimeOffsetConverter : JsonConverter<DateTimeOffset>
+{
+    /// <inheritdoc />
+    public override DateTimeOffset Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        => reader.TokenType == JsonTokenType.Null
```
baywet, line 15: Same comment about the condition being reversed

tests/serialization/json/Converters/JsonDateTimeOffsetConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+            ? new DateTimeOffset()
+            : ReadInternal(ref reader);
+
+    /// <inheritdoc />
+    public override void Write(Utf8JsonWriter writer, DateTimeOffset value, JsonSerializerOptions options)
+        => WriteInternal(writer, value);
+
+    private static DateTimeOffset ReadInternal(ref Utf8JsonReader reader)
+        => DateTimeOffset.ParseExact(reader.GetString()!, "dd__MM__yyyyTHH_mm_ssZ", CultureInfo.InvariantCulture);
```
baywet, line 24: Same comment about the bang operator

tests/serialization/json/Converters/JsonDateTimeOffsetConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    private static void WriteInternal(Utf8JsonWriter writer, DateTimeOffset value)
+        => writer.WriteStringValue(value.ToString("dd__MM__yyyyTHH_mm_ssZ"));
```
baywet, line 27: Maybe we could make the template a constant

tests/serialization/json/Converters/JsonDateConverter.cs
MartinM85: ok, I can simply call value.ToString

baywet: I'm not saying this is wrong i'm just trying to understand the context for it

tests/serialization/json/Converters/JsonDateConverter.cs
MartinM85: Seems to me correct, TokenType is string and in that case ReadInternal is called

tests/serialization/json/Converters/JsonDateConverter.cs
MartinM85: In that case it doesn't matter, because JsonSerializationWriter won't use custom converter, same for Time

baywet (approved):
Thanks for the great work here!
@andrueastman for final review, merge and publish

andrueastman (approved):

PR #303, opened 2024-07-22
fix: automatic decompression of response content

### LINKED ISSUE #302: compress handler name should be decompression handler
related https://github.com/microsoft/kiota/issues/598
It should be named DecompressionHandler since what it does is to decompress responses.
Also, since the client supports automatic decompression, there's not much value in having it in the first place

baywet:
fixes #302


andrueastman (approved):

baywet:
@andrueastman we'll need to remove the obsolete handler from the graph core client factory and bump the required http version as a follow up. Can you take care of it please?

PR #307, opened 2024-07-25
Add support to JsonSerializationWriter.WriteNonParsableObjectValue for anonymous objects.

### LINKED ISSUE #306: JsonSerializationWriter.WriteNonParsableObjectValue does not support anonymous objects
I was passing values into the AdditionalData property on the type generated in my Kiota client and I noticed that anonymous object values were being passed as empty objects.

```
{ "myAdditionalDataProperty": { } }
```

From debugging I can see that here https://github.com/microsoft/kiota-dotnet/blob/42ea74edf45d473462dd050661b1972ce14c15e1/src/serialization/json/JsonSerializationWriter.cs#L447 T comes through as "Object" so get properties returns no results. 

No error is thrown, the properties just never show up. It's possible this affects other types but I only tested with anonymous objects.

greg-ww:
Fixes #306.

### DIFF
--- a/src/serialization/json/JsonSerializationWriter.cs
+++ b/src/serialization/json/JsonSerializationWriter.cs
@@ -444,7 +444,7 @@ namespace Microsoft.Kiota.Serialization.Json
             if(value == null)
                 writer.WriteNullValue();
             else
-                foreach(var oProp in typeof(T).GetProperties())
+                foreach(var oProp in value.GetType().GetProperties())
                     WriteAnyValue(oProp.Name, oProp.GetValue(value));
             writer.WriteEndObject();
         }
--- a/src/serialization/json/JsonSerializationWriter.cs
+++ b/src/serialization/json/JsonSerializationWriter.cs
@@ -380,7 +380,7 @@ namespace Microsoft.Kiota.Serialization.Json
                 foreach(DictionaryEntry entry in values)
                 {
                     if(entry.Key is not string keyStr)
-                        throw new InvalidOperationException($"error serialization additional data value with key {key}, unsupported type {values.GetType()}");
+                        throw new InvalidOperationException($"Error serializing dictionary value with key {key}, only string keyed dictionaries are supported.");
                     WriteAnyValue(keyStr, entry.Value);
                 }
                 writer.WriteEndObject();
@@ -549,7 +549,7 @@ namespace Microsoft.Kiota.Serialization.Json
                     WriteNullValue(key);
                     break;
                 default:
-                    throw new InvalidOperationException($"error serialization additional data value with key {key}, unknown type {value?.GetType()}");
+                    throw new InvalidOperationException($"Error serializing additional data value with key {key}, unknown type {value?.GetType()}");
             }
         }
 
--- a/src/serialization/json/JsonSerializationWriter.cs
+++ b/src/serialization/json/JsonSerializationWriter.cs
@@ -375,14 +375,13 @@ namespace Microsoft.Kiota.Serialization.Json
             {
                 if(!string.IsNullOrEmpty(key))
                     writer.WritePropertyName(key!);
-                var type = values.GetType();
-                if(!type.IsGenericType || type.GetGenericArguments()[0] != typeof(string))
-                    throw new InvalidOperationException($"error serialization additional data value with key {key}, unsupported type {type}");
 
                 writer.WriteStartObject();
                 foreach(DictionaryEntry entry in values)
                 {
-                    WriteAnyValue((string)entry.Key, entry.Value);
+                    if(entry.Key is not string keyStr)
+                        throw new InvalidOperationException($"error serialization additional data value with key {key}, unsupported type {values.GetType()}");
+                    WriteAnyValue(keyStr, entry.Value);
                 }
                 writer.WriteEndObject();
             }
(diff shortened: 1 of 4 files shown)


src/serialization/json/JsonSerializationWriter.cs
```diff
@@ -444,7 +472,7 @@ private void WriteNonParsableObjectValue<T>(string? key, T value)
             if(value == null)
                 writer.WriteNullValue();
             else
-                foreach(var oProp in typeof(T).GetProperties())
+                foreach(var oProp in value.GetType().GetProperties())
```
baywet, line 475: why this change?

src/serialization/json/JsonSerializationWriter.cs
```diff
@@ -360,6 +361,33 @@ public void WriteCollectionOfPrimitiveValues<T>(string? key, IEnumerable<T>? val
...
+                    writer.WritePropertyName(key!);
+                var type = values.GetType();
+                if(!type.IsGenericType || type.GetGenericArguments()[0] != typeof(string))
+                    throw new InvalidOperationException($"error serialization additional data value with key {key}, unsupported type {type}");
+
+                writer.WriteStartObject();
+                foreach(DictionaryEntry entry in values)
+                {
+                    WriteAnyValue((string)entry.Key, entry.Value);
+                }
```
baywet, line 386:
```suggestion
                writer.WriteStartObject();
                foreach(DictionaryEntry entry in values)
                {
                    if (entry.Key is not string keyStr)
                        throw new InvalidOperationException($"error serialization additional data value with key {key}, unsupported type {type}");
                    WriteAnyValue(keyStr, entry.Value);
                }
```

While functional, advanced type reflection has a high runtime cost and is going to trip AOT/trimming

baywet (changes requested):
Thanks for the contributions!
A couple of remarks, also please sign the CLA (see the bot comment)

greg-ww:
@microsoft-github-policy-service agree company="Civitai"

src/serialization/json/JsonSerializationWriter.cs
greg-ww: I believe this is covered in the issue #306, but typeof(T) comes back as object and therefore no properties are ever returned by GetProperties(). For a clear demonstration of the issue you can revert this change and run the WriteSampleObjectValue test I modified.

src/serialization/json/JsonSerializationWriter.cs
greg-ww: Happy to iterate on this.
Is there something specific you would rather I avoid?

baywet: yes, we should avoid using reflection as much as possible here since it throws AOT/trimming off and is costly at runtime.
I believe the suggestion I've made is functionally equivalent to the initial proposal here.

greg-ww: Ah sorry I missed the suggestion, yeah that works for me.

greg-ww: We still need to use reflection for the error message but this is inline with the WriteAnyValue so I assume that's fine.

greg-ww: I will look at adding a test for the exception too.

src/serialization/json/JsonSerializationWriter.cs
baywet: Thanks for the additional information. I'd like @andrueastman to provide his input here. I think this should be fine because of the annotation, but would like a second opinion here.

andrueastman: Yeah, this change is okay with me really.
The codebase was always using `value.GetType()` but the use of `typeof(T)` snuck in recently in this PR https://github.com/microsoft/kiota-dotnet/pull/291/files when we were trying to iron out AOT/trimming warnings with the test project. So, this is probably looking to fix a regression that we didn't catch.

andrueastman (commented):
Thanks for the contribution @greg-ww 

Any chance you can do the following as well?

- Run `dotnet format` to fix linting errors
- Bump the patch version at https://github.com/microsoft/kiota-dotnet/blob/1eca71b6e4c86f8da5db1c420d1a3ba2cd71ab06/Directory.Build.props#L4
- Add a changelog entry at https://github.com/microsoft/kiota-dotnet/blob/main/CHANGELOG.md

greg-ww:
Will try to implement this feedback today.

src/serialization/json/JsonSerializationWriter.cs
```diff
@@ -360,6 +361,32 @@ public void WriteCollectionOfPrimitiveValues<T>(string? key, IEnumerable<T>? val
             }
         }
         /// <summary>
+        /// Writes the specified dictionary to the stream with an optional given key.
+        /// </summary>
+        /// <param name="key">The key to be used for the written value. May be null.</param>
+        /// <param name="values">The dictionary of values to be written.</param>
+#if NET5_0_OR_GREATER
+        public void WriteDictionaryValue<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)] T>(string? key, T values) where T : IDictionary
```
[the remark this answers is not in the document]
greg-ww: Yeah fair enough, will do.

greg-ww:
@andrueastman I made the requested modifications, let me know if I missed anything.

andrueastman (approved):
Looks good to me @greg-ww 

Just need a final review from @baywet before we merge.

baywet (approved):

PR #311, opened 2024-07-31
Improving serialization

### LINKED ISSUE #310: Idea: Use Kiota generated models for serialization
I was wondering if I could use the models together with the used `JsonSerializerOptions` to read/write json.

The use case is that I would want to save some results of the Graph API, to a folder. Talking *conditional access policies* here, but any model will do. And I don't want to do any customizations, just save the object exactly as it would be transferred to and from the api.

### pseudo code

```csharp
var client = new GraphServiceClient(deviceCodeCredential, scopes: ["User.Read"]);

// Get something from the Kiota client
var user = await client.Me.GetAsync();

// This does not work
Microsoft.Kiota.Serialization.Json.JsonSerializationWriter writer = new Microsoft.Kiota.Serialization.Json.JsonSerializationWriter(new Microsoft.Kiota.Serialization.Json.KiotaJsonSerializationContext());
user!.Serialize(writer);
var stream = writer.GetSerializedContent();
// maybe this would be nice!
var userStream = user!.SerializeAsJsonStream(); // extension method?

// Save the stream to a file
using var fileStream = File.Create(filename);
stream.Seek(0, SeekOrigin.Begin);
stream.CopyTo(fileStream);

// and on the other hand, I want to read it back
using var fileStream2 = File.OpenRead(filename);
Microsoft.Graph.Models.User userFromDisk = ... // How to deserialize this?
```

### Discussion

This is not really an issue, but this repo doesn't have discussions enabled.

svrooij:
Fixed #310


src/abstractions/serialization/KiotaJsonSerializer.Serialization.cs
```diff
@@ -20,9 +20,21 @@ public static partial class KiotaJsonSerializer
     /// </summary>
     /// <typeparam name="T">Type of the object to serialize</typeparam>
     /// <param name="value">The object to serialize.</param>
+    /// <param name="useBackingStore">Backing store keeps track of changes, setting this to false will give you the full item.</param>
```
baywet, line 23: could we name this parameter something else like "serializeOnlyChangedValues" ?

src/abstractions/serialization/KiotaJsonSerializer.Serialization.cs
```diff
@@ -20,9 +20,21 @@ public static partial class KiotaJsonSerializer
...
+
+    /// <summary>
+    /// Serializes the specified object as a string based JSON stream.
+    /// </summary>
+    /// <typeparam name="T">The type of the value to serialize.</typeparam>
+    /// <param name="value">The object to serialize.</param>
+    /// <param name="useBackingStore">A flag indicating whether to use a backing store for serialization.</param>
+    /// <returns>A <see cref="Stream"/> containing the serialized JSON data.</returns>
+
+    public static Stream SerializeAsJsonStream<T>(this T value, bool useBackingStore = true) where T : IParsable
```
baywet, line 36: we need to replicate the change for all other public methods (collections, etc...)

src/abstractions/serialization/KiotaSerializer.Serialization.cs
```diff
@@ -25,11 +25,12 @@ public static partial class KiotaSerializer
...
     /// <param name="contentType">Content type to serialize the object to </param>
     /// <param name="value">The object to serialize.</param>
+    /// <param name="useBackingStore">Backing store keeps track of changes, setting this to false will give you the full item.</param>
     /// <returns>The serialized representation as a stream.</returns>
-    public static Stream SerializeAsStream<T>(string contentType, T value) where T : IParsable
+    public static Stream SerializeAsStream<T>(string contentType, T value, bool useBackingStore = true) where T : IParsable
     {
         using var writer = GetSerializationWriter(contentType, value);
-        writer.WriteObjectValue(string.Empty, value);
+        writer.WriteObjectValue(useBackingStore ? string.Empty : null, value);
```
baywet, line 33: I'm not sure I follow how switching between string empty and null impacts the backing store behaviour here? Those seem like two different concerns.

baywet (changes requested):
Thanks for starting the work here.

src/abstractions/serialization/KiotaSerializer.Serialization.cs
svrooij: I tried this code, mind the `null` instead of `string.empty`. and it gives me back the entire item.

```csharp

    public static Stream? SerializeAsJsonStream<T>(this T? input) where T : IParsable
    {
        if (input == null) return null;
        var writer = new Microsoft.Kiota.Serialization.Json.JsonSerializationWriter(new Microsoft.Kiota.Serialization.Json.KiotaJsonSerializationContext());
        writer.WriteObjectValue(null, input);
        return writer.GetSerializedContent();
    }
```

src/abstractions/serialization/KiotaJsonSerializer.Serialization.cs
svrooij: Yes

src/abstractions/serialization/KiotaJsonSerializer.Serialization.cs
svrooij: You mean you want the extension method also on collections?
And maybe also on the string versions?

src/abstractions/serialization/KiotaSerializer.Serialization.cs
baywet: the null key is not what's changing things here. It's the fact that the writer does not have the proxies registered for the backing store. (you're creating a new writer) so it reverts to the default behaviour (serialize everything)
That's actually a better solution than what I was suggesting (flip the backing store flags all the way across the tree) but we need to make sure we implement it in a way that does not hardcode things with a specific writer.
One way to do that would be to assign null to the events receivers of the writer when the parameter is true.

svrooij: Oops, I was to quick.

This abstraction library has no clue about the actual serializer (it's in the abstraction library). Do you mean:

```csharp
        using var writer = GetSerializationWriter(contentType, value);
        if (!serializeOnlyChangedValues) {
            writer.OnAfterObjectSerialization = null;
            writer.OnBeforeObjectSerialization = null;
            writer.OnStartObjectSerialization = null;
        }
        writer.WriteObjectValue(string.Empty, value);
```

andrueastman: I think this may be better. Resetting the hooks may cause unintentional side effects later on if the model is being re-used and one expects them to still  be there or added them in a custom way. 

This enables us to target situation where the backing store is not returning all the values.....
```cs
        if(!serializeOnlyChangedValues && value is IBackedModel backedModel)
        {
            backedModel.BackingStore.InitializationCompleted = false;// reset the initialization completed flag to ensure all properties are serialized
        }
```

svrooij: Great idea @andrueastman, have any idea how I would do that on a collection?

baywet: right, but if we choose this approach (what I initially had suggested), we'll need to do it for the whole tree.
we'll also need to reset the value back after the serialization.

maybe mixing both solutions would be the better approach here if a hook is registered, register an additional one that flips things up.
This way we don't impact the original behaviour (which might or might not be backing store related) and fix the issue WITHOUT having to go through the whole tree ourselves.
Thoughts?

svrooij: If we want [this](https://github.com/microsoft/kiota-dotnet/pull/311#discussion_r1698549272) the extension methods have to go into the *Kiota.Serialization.Json* library, because that is where it's known.

The solution @andrueastman is suggesting is applied on the code above. No sure if the collection part is working though.

andrueastman: > This way we don't impact the original behavior (which might or might not be backing store related) and fix the issue WITHOUT having to go through the whole tree ourselves.

I agree with @baywet here. If we want to reset the flag/return the store to original state, we should probably reset the `ReturnOnlyChangedValues` flag and NOT the `InitializationCompleted` (as this will clear the store state).

Another idea we could probably do is add a public boolean property to the [BackingStoreSerializationWriterProxyFactory](https://github.com/microsoft/kiota-dotnet/blob/main/src/abstractions/store/BackingStoreSerializationWriterProxyFactory.cs#L12) that defaults to true. Then the hooks would use it to do their current actions. Otherwise allow all properties to be serialized by doing the opposite action.

If that works, we would simply need to update the code whenever [GetSerializationWriter](https://github.com/microsoft/kiota-dotnet/blob/bc0404d400e977e41c0dc277a845d6bf977ef44b/src/abstractions/serialization/KiotaSerializer.Serialization.cs#L115) is called to do something like this.

```cs
        if(!serializeOnlyChangedValues && writer is BackingStoreSerializationWriterProxyFactory)
            BackingStoreSerializationWriterProxyFactory.ReturnOnlyChangedValues = false; //enable writing all props

        writer.WriteObjectValue(string.Empty, value);

        if(!serializeOnlyChangedValues && writer is BackingStoreSerializationWriterProxyFactory)
            BackingStoreSerializationWriterProxyFactory.ReturnOnlyChangedValues = true; // reset to original state
```

src/abstractions/serialization/KiotaJsonSerializer.Serialization.cs
```diff
@@ -20,9 +20,21 @@ public static partial class KiotaJsonSerializer
...
+
+    /// <summary>
+    /// Serializes the specified object as a string based JSON stream.
+    /// </summary>
+    /// <typeparam name="T">The type of the value to serialize.</typeparam>
+    /// <param name="value">The object to serialize.</param>
+    /// <param name="serializeOnlyChangedValues">By default you'll only get the changed properties</param>
+    /// <returns>A <see cref="Stream"/> containing the serialized JSON data.</returns>
+
+    public static Stream SerializeAsJsonStream<T>(this T value, bool serializeOnlyChangedValues = true) where T : IParsable
```
baywet, line 36: I originally did not catch the fact this was an extension method. 
Should we roll this change back? Because it's going to provide a different experience between the different data types. Also with other languages.

src/abstractions/serialization/KiotaJsonSerializer.Serialization.cs
baywet: I meant the optional parameter

src/abstractions/serialization/KiotaSerializer.Serialization.cs
```diff
@@ -25,12 +27,25 @@ public static partial class KiotaSerializer
     /// <typeparam name="T">Type of the object to serialize</typeparam>
     /// <param name="contentType">Content type to serialize the object to </param>
     /// <param name="value">The object to serialize.</param>
+    /// <param name="serializeOnlyChangedValues">By default you'll only get the changed properties.</param>
     /// <returns>The serialized representation as a stream.</returns>
-    public static Stream SerializeAsStream<T>(string contentType, T value) where T : IParsable
+    public static Stream SerializeAsStream<T>(string contentType, T value, bool serializeOnlyChangedValues = true) where T : IParsable
     {
+        bool restoreInitializationCompleted = false;
+        if(!serializeOnlyChangedValues && value is IBackedModel backedModel)
```
baywet, line 35: That's only going to work for the root object. 
I think that instead of doing that we should update the event listeners to switch the flag and call into the existing receiver instead.

src/abstractions/serialization/KiotaJsonSerializer.Serialization.cs
svrooij: Hmm... I explicitly choose it to be an extension method, this will help discovery.

That other languages don't support extension methods, resulting in not using extension methods every where, seems a political decision (which I would strongly disagree with).
At best this difference would convince developers to start thinking about switching, and at worst developers could be disappointed in the languages that does not support extension methods.

baywet: I don't think we want to be in the business of convincing people to switch to dotnet.
Does the IEnumerable<IParsable> extension method work for List<Parsable> or Parsable[] for example? (I have had surprises with generics and covariance in the past)
Should they be under a different namespace to avoid potential collision/confusion? (.Extensions)

src/abstractions/serialization/ISerializationWriterFactory.cs
```diff
@@ -17,7 +17,8 @@ public interface ISerializationWriterFactory
         ///     Creates a new <see cref="ISerializationWriter" /> instance for the given content type.
         /// </summary>
         /// <param name="contentType">The content type for which a serialization writer should be created.</param>
+        /// <param name="serializeOnlyChangedValues">By default a backing store is used, and you'll only get changed properties</param>
         /// <returns>A new <see cref="ISerializationWriter" /> instance for the given content type.</returns>
-        ISerializationWriter GetSerializationWriter(string contentType);
+        ISerializationWriter GetSerializationWriter(string contentType, bool serializeOnlyChangedValues = true);
```
baywet, line 22: We can't make this change as it'll be source breaking

baywet (changes requested):

src/abstractions/serialization/KiotaJsonSerializer.Serialization.cs
svrooij: It will work on list not on `IParsable[]` because it does not see an array as a generic Ienumerable (last time I checked)

src/abstractions/serialization/KiotaSerializer.Serialization.cs
svrooij: @andrueastman You're not able to access properties inside an action you create in the base constructor.

But I've slightly modified the [SerializationWriterFactoryRegistration](https://github.com/microsoft/kiota-dotnet/pull/311/files#diff-f305f58e4f7d9e3de38c6908dba355140eabdd5970ff388c40677291adbf64d6) to enable this change

svrooij:
It was a bit of pickle, but I think this is a pretty nice solution. Did I just notice that the backing store stuff if not referenced in any tests?

src/abstractions/serialization/KiotaJsonSerializer.Serialization.cs
baywet: Thank you for the additional information. I think this is what is bothering me the most: the differences in experiences between the different types:
- Parsable -> stream: extension works
- IEnumerable of Parsable -> stream: extension works
- Parsable[] -> stream: extension doesn't work
- Parsable -> string: no extension
- IEnumerable of Parsable -> string: no extension
- Parsable[] -> string: no extension

Also, why would we have extension methods only for Json?

Now, if we wanted to cover everything, we'd need lots of additional methods. At which point I still believe they should be in their own file/namespace.
The alternative being: switching all the existing methods to ALSO be extension methods, they should still work as a static method.
Both approaches won't solve the array aspect, but that could be solved by adding additional extension methods to do the mapping.

Thoughts?

src/abstractions/serialization/SerializationWriterProxyFactory.cs
```diff
@@ -15,7 +15,11 @@ public class SerializationWriterProxyFactory : ISerializationWriterFactory
         /// The valid content type for the <see cref="SerializationWriterProxyFactory"/>
         /// </summary>
         public string ValidContentType { get { return _concrete.ValidContentType; } }
-        private readonly ISerializationWriterFactory _concrete;
+
+        /// <summary>
+        /// The concrete factory to wrap.
+        /// </summary>
+        protected readonly ISerializationWriterFactory _concrete;
```
baywet, line 22: This should follow the naming conventions: `Concrete` and while at it, we might want to give it a better name now that it's exposed. Like "RealSerializationWriterFactory"

src/abstractions/serialization/SerializationWriterFactoryRegistry.cs
```diff
@@ -51,5 +51,45 @@ public ISerializationWriter GetSerializationWriter(string contentType)
...
+        }
+
+        /// <summary>
+        /// Get the relevant <see cref="ISerializationWriterFactory"/> instance for the given content type
+        /// </summary>
+        /// <param name="contentType">The content type in use</param>
+        /// <returns></returns>
+        /// <exception cref="ArgumentNullException"></exception>
+        /// <exception cref="InvalidOperationException"></exception>
+        private ISerializationWriterFactory GetSerializationWriterFactory(string contentType)
```
baywet, line 79: I believe the GetSerializationWriter could use this method and return the writer instead of duplicating the code.

svrooij: Good point

src/abstractions/serialization/SerializationWriterProxyFactory.cs
svrooij: `protected readonly ISerializationWriterFactory ProxiedSerializationWriterFactory;`

src/abstractions/serialization/KiotaJsonSerializer.Serialization.cs
svrooij: In the latest update I put them in their respective projects.

If you're like me and did remove some of the serializers, you won't need those packages included as well. If you would then use the `KiotaJsonSerializer`, it would throw an exception. In the new case you'll need the actual using to `Microsoft.Kiota.Serialization.Json` and each IParsable will have `SerializeAsStream` and `SerializeAsJsonStringAsync` (and arrays and ienumerables)

src/abstractions/serialization/SerializationWriterFactoryRegistry.cs
```diff
@@ -26,30 +26,63 @@ public string ValidContentType
...
+        /// <summary>
+        /// Get the relevant <see cref="ISerializationWriter"/> instance for the given content type
+        /// </summary>
+        /// <param name="contentType">The content type in use</param>
+        /// <param name="serializeOnlyChangedValues">If <see langword="true"/> will only return changed values, otherwise will return the full object </param>
+        /// <returns>A <see cref="ISerializationWriter"/> instance to parse the content</returns>
+        public ISerializationWriter GetSerializationWriter(string contentType, bool serializeOnlyChangedValues)
+        {
+            var factory = GetSerializationWriterFactory(contentType, out string actualContentType);
+            if(!serializeOnlyChangedValues && factory is Store.BackingStoreSerializationWriterProxyFactory backingStoreFactory)
```
andrueastman, line 52: 👍🏼

svrooij:
I'm out of the office this week, and see that the formatting failed.

I checked the box that I gave the contributors access to the branch. So if anybody wants to run `dotnet format` on my code, please do 🤩

src/abstractions/serialization/KiotaSerializer.Serialization.cs
andrueastman: > @andrueastman You're not able to access properties inside an action you create in the base constructor.

I see that now. The suggested approach look great.

src/abstractions/serialization/KiotaJsonSerializer.Serialization.cs
andrueastman: The addition of the methods (at least for Json) introduces some duplication with the existing helpers from Abstractions (I think we'd at least need to create an issue to come clean this up later on the next major version if we go this route). 
I do like the suggestion by @baywet to convert the helpers to extension methods (we can still have consistency on other languages and call them the same way as now).

> If you're like me and did remove some of the serializers, you won't need those packages included as well. If you would then use the KiotaJsonSerializer it would throw an exception.

The problem here would be registration of the custom serializer and not where the code would reside. If the custom serilizer is registered, it should not matter where we are at.

svrooij: We could create them as extension methods. But those need the object you want to convert to be the first parameter. That would be a source breaking change.

baywet: in the interest of getting this pull request merged can we please:
- create another issue to bring the extension methods
- remove any extension method from this PR

svrooij: @baywet done

baywet (approved):
@andrueastman for final review and merge.
I believe we'll need a version bump/changelog entry as well.

baywet (dismissed):

baywet (approved):

Issue #284, opened 2024-06-26
Enum serialization should not be changing casing


baywet:
https://github.com/microsoft/kiota-serialization-form-dotnet/blob/50d7a29ac768c2f3c0047d8909f12c2df623a30f/src/FormSerializationWriter.cs#L235

(x3 instances)

should be using the enum member attribute instead like here

https://github.com/microsoft/kiota-serialization-json-dotnet/blob/c9c3fe221f52d26322d6b1c8bb562444fdacb4e4/src/JsonSerializationWriter.cs#L539

Issue #281, opened 2024-07-04
Pass relevant `JsonWriterOptions` from the `KiotaJsonSerializationContext.Options` to the `Utf8JsonWriter` when writing json to enable customization.


andrueastman:
Related to discussion at https://github.com/microsoft/kiota/discussions/4898#discussioncomment-9949732

Trying out the code below, the Encode settings seem to have no effect on the serialized payload.

```cs
        var jsonSerializationOptions = new JsonSerializerOptions
        {
            WriteIndented = true,
            Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping
        };
        var kiotaJsonSerializationContext = new KiotaJsonSerializationContext(jsonSerializationOptions); // create serialization context with custom options
        SerializationWriterFactoryRegistry.DefaultInstance.ContentTypeAssociatedFactories["application/json"] = new JsonSerializationWriterFactory(kiotaJsonSerializationContext); // replace the json serializer with one with custom options

        var requestJson = await KiotaJsonSerializer.SerializeAsStringAsync(request);
```

This is because, the `JsonSerializationWriter` initializes a `Utf8JsonWriter` instance without propagating relelvant options to the `JsonWriterOptions` parameter in the constructor here. 

https://github.com/microsoft/kiota-serialization-json-dotnet/blob/046356f4e43767b12cfa7b6d223671d83e63309c/src/JsonSerializationWriter.cs#L51

The line should probably be updated to something as below and relevant tests added.

```cs
writer = new Utf8JsonWriter(_stream, new JsonWriterOptions {
    Encoder = kiotaJsonSerializationContext.Options.Encoder,
    Indented = kiotaJsonSerializationContext.Options.WriteIndented
});
```

PR #312, opened 2024-08-01
Aligns Enum serialization across packages and pass relevant JsonWriterOptions from the KiotaJsonSerializationContext

andrueastman:
Fixes https://github.com/microsoft/kiota-dotnet/issues/281
Fixes https://github.com/microsoft/kiota-dotnet/issues/284

This PR 
- Cleans up enum serialization to read from attributes for form and text serialization. This uses the shared helper to provide consistent results. Tests updated.
- Pass relevant `JsonWriterOptions` from the `KiotaJsonSerializationContext.Options` to the `Utf8JsonWriter` when writing json to enable customization for the writer.

### DIFF
--- a/src/abstractions/Helpers/EnumHelpers.cs
+++ b/src/abstractions/Helpers/EnumHelpers.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Reflection;
 using System.Runtime.Serialization;
+using Microsoft.Kiota.Abstractions.Extensions;
 
 #if NET5_0_OR_GREATER
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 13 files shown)


src/abstractions/Helpers/EnumHelpers.cs
```diff
@@ -154,5 +155,29 @@ private static bool TryGetFieldValueName(Type type, string rawValue, out string
...
+        {
+            var type = typeof(T);
+
+            if(Enum.GetName(type, value) is not { } name)
+                throw new ArgumentException($"Invalid Enum value {value} for enum of type {type}");
+
+            if(type.GetField(name)?.GetCustomAttribute<EnumMemberAttribute>() is { } attribute)
+                return attribute.Value;
+
+            return name.ToFirstCharacterLowerCase();
```
baywet, line 180: Oversight?

andrueastman: Thanks for highlighting this. I am hesitant to drop this one as this only happens when there is no `EnumMember` attribute where it tries to camelCase the property. 
Given that the convention is to have the enum members to start with uppercase in C#, this means that member will typically be serialized in pascal case which is not an expectation we've set in previous releases.

baywet: this is a thing only dotnet is doing today.
Assuming we don't have special casing in the generation for dotnet to add/remove the enum member attributes (i.e. it follows the same logic as the other languages that are upper camel cased) we should be fine.
But yes, we should make sure the generation is aligned before we make that change.

baywet (approved):
thanks for making the changes!

PR #326, opened 2024-08-12
fix: performance issue with the backing store

baywet:
fixes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/2563


andrueastman (approved):
👍🏼

baywet:
@grippstick thanks for the follow up! Authored #328

PR #329, opened 2024-08-15
Enable nullable context in all test projects

### LINKED ISSUE #323: Enable `nullable` context in all test projects.
Source of the project support NRT. However some of the test projects have this disabled. This should be aligned to clean up the code base.

Example at https://github.com/microsoft/kiota-dotnet/blob/8be19ce7ecaa92c001101f14e6acf6aecd57c47d/tests/abstractions/Microsoft.Kiota.Abstractions.Tests.csproj#L6

bm0niz:
Fixes #323 

Remarks:

- `IRequestAdapter.ConvertToNativeRequestAsync`: Couldn't find a path in the implementation that returns a nullable *HttpRequestMessage*. Please check.
- `<Nullable>enable</Nullable>` prop now flows from the root.

### DIFF
--- a/src/abstractions/IRequestAdapter.cs
+++ b/src/abstractions/IRequestAdapter.cs
@@ -86,6 +86,6 @@ namespace Microsoft.Kiota.Abstractions
         /// <param name="requestInfo">The RequestInformation object to use for the HTTP request.</param>
         /// <param name="cancellationToken">The <see cref="CancellationToken"/> to use for cancelling the requests.</param>
         /// <returns>The native HTTP request.</returns>
-        Task<T> ConvertToNativeRequestAsync<T>(RequestInformation requestInfo, CancellationToken cancellationToken = default);
+        Task<T?> ConvertToNativeRequestAsync<T>(RequestInformation requestInfo, CancellationToken cancellationToken = default);
     }
 }
--- a/src/abstractions/IRequestAdapter.cs
+++ b/src/abstractions/IRequestAdapter.cs
@@ -86,6 +86,6 @@ namespace Microsoft.Kiota.Abstractions
         /// <param name="requestInfo">The RequestInformation object to use for the HTTP request.</param>
         /// <param name="cancellationToken">The <see cref="CancellationToken"/> to use for cancelling the requests.</param>
         /// <returns>The native HTTP request.</returns>
-        Task<T?> ConvertToNativeRequestAsync<T>(RequestInformation requestInfo, CancellationToken cancellationToken = default);
+        Task<T> ConvertToNativeRequestAsync<T>(RequestInformation requestInfo, CancellationToken cancellationToken = default);
     }
 }
--- a/tests/Directory.Build.props
+++ b/tests/Directory.Build.props
@@ -3,7 +3,6 @@
 	<PropertyGroup>
 		<TargetFrameworks>net8.0;net462</TargetFrameworks>
 		<IsTestProject>true</IsTestProject>
-		<Nullable>disable</Nullable>
 		<ImplicitUsings>true</ImplicitUsings>
 		<LangVersion>latest</LangVersion>
 	</PropertyGroup>
--- a/tests/Directory.Build.props
+++ b/tests/Directory.Build.props
@@ -5,5 +5,6 @@
 		<IsTestProject>true</IsTestProject>
 		<ImplicitUsings>true</ImplicitUsings>
 		<LangVersion>latest</LangVersion>
+		<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
 	</PropertyGroup>
 </Project>
--- a/tests/http/httpClient/Extensions/HttpRequestMessageExtensionsTests.cs
+++ b/tests/http/httpClient/Extensions/HttpRequestMessageExtensionsTests.cs
@@ -72,7 +72,7 @@ namespace Microsoft.Kiota.Http.HttpClientLibrary.Tests.Extensions
 
             var clonedRequest = await originalRequest.CloneAsync();
             var originalContents = await originalRequest.Content.ReadAsStringAsync();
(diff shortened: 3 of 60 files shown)


bm0niz:

tests/serialization/text/Microsoft.Kiota.Serialization.Text.Tests.csproj
```diff
@@ -1,9 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <Nullable>enable</Nullable> <!-- This test project supports NRT other projects need cleanup as outlined in https://github.com/microsoft/kiota-dotnet/issues/323 -->
```
andrueastman, line 4: 👍🏼

tests/Directory.Build.props
```diff
@@ -3,7 +3,6 @@
 	<PropertyGroup>
 		<TargetFrameworks>net8.0;net462</TargetFrameworks>
 		<IsTestProject>true</IsTestProject>
-		<Nullable>disable</Nullable>
```
andrueastman, line 6: Could we also add `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` so that any future warnings now break the build?

src/abstractions/IRequestAdapter.cs
```diff
@@ -86,6 +86,6 @@ public interface IRequestAdapter
         /// <param name="requestInfo">The RequestInformation object to use for the HTTP request.</param>
         /// <param name="cancellationToken">The <see cref="CancellationToken"/> to use for cancelling the requests.</param>
         /// <returns>The native HTTP request.</returns>
-        Task<T?> ConvertToNativeRequestAsync<T>(RequestInformation requestInfo, CancellationToken cancellationToken = default);
```
andrueastman, line 89: I think we could probably revert this for now if we can. As `IRequestAdapter` is a public interface, an alternative implementation (not necessarily implemented by us) may choose to return the null type as much as our implementation here does not. 
Also tagging @baywet to weigh in here.

tests/serialization/json/JsonParseNodeTests.cs
```diff
@@ -175,7 +175,7 @@ public void GetsChildNodeAndGetCollectionOfPrimitiveValuesFromJsonParseNode()
             var rootParseNode = new JsonParseNode(jsonDocument.RootElement);
             // Act to get business phones list
             var phonesListChildNode = rootParseNode.GetChildNode("businessPhones");
-            var phonesList = phonesListChildNode.GetCollectionOfPrimitiveValues<string>().ToArray();
+            var phonesList = phonesListChildNode!.GetCollectionOfPrimitiveValues<string>().ToArray();
```
andrueastman, line 178:
```suggestion
            Assert.NotNull(phonesListChildNode);
            var phonesList = phonesListChildNode.GetCollectionOfPrimitiveValues<string>().ToArray();
```

tests/trimming/Microsoft.Kiota.Trimming.Validation/Microsoft.Kiota.Trimming.Validation.csproj
bm0niz, line 1: I think this should also be restored since this project doesn't inherit root props

tests/Directory.Build.props
bm0niz: [Added](https://github.com/microsoft/kiota-dotnet/pull/329/commits/2b5ac5fa883759f24f4a9b1cec3b0fd8f1ded82b)

src/abstractions/IRequestAdapter.cs
bm0niz: 👍 👍 
Changes reverted in the meantime

tests/http/httpClient/Extensions/HttpRequestMessageExtensionsTests.cs
```diff
@@ -34,7 +34,7 @@ public async Task GetRequestOptionCanExtractRequestOptionFromHttpRequestMessage(
             };
             requestInfo.AddRequestOptions(new IRequestOption[] { redirectHandlerOption });
             // Act and get a request message
-            var requestMessage = await requestAdapter.ConvertToNativeRequestAsync<HttpRequestMessage>(requestInfo);
+            var requestMessage = (await requestAdapter.ConvertToNativeRequestAsync<HttpRequestMessage>(requestInfo))!;
```
baywet, line 37: could we instead assert not null on the request message? It'll probably make the error message more intelligible if the behaviour changes.

baywet (changes requested):
Thanks for the contribution!

src/abstractions/IRequestAdapter.cs
baywet: In general I'd rather avoid changing the interfaces here, so thank you for reverting. In this specific instance, I'm not sure "allowing" implementers to return null makes sense. They probably should return something or throw to avoid derailing callers. Andrew: do you remember the context here?

tests/http/httpClient/Extensions/HttpRequestMessageExtensionsTests.cs
bm0niz: Done @baywet - check please [743e8b8](https://github.com/microsoft/kiota-dotnet/pull/329/commits/743e8b8aeeb54de5ff6e43d2b2d2fc94178366da)

src/abstractions/IRequestAdapter.cs
andrueastman: I think this came in as part of the of the NRT work with all other methods in the interface returning nullable types. At that time the repos were separate and this may be clearer now with the monorepo with the implementation and the interface together. If this was added today, we'd probably not do this as you suggest.

baywet (changes requested):

baywet (approved):

PR #341, opened 2024-08-21
fix: missing cancellation token

baywet:


src/abstractions/serialization/ParseNodeProxyFactory.cs
```diff
@@ -27,7 +27,7 @@ public abstract class ParseNodeProxyFactory : IAsyncParseNodeFactory
         /// <param name="concrete">The concrete factory to wrap.</param>
         /// <param name="onBefore">The callback to invoke before the deserialization of any model object.</param>
         /// <param name="onAfter">The callback to invoke after the deserialization of any model object.</param>
-        public ParseNodeProxyFactory(IParseNodeFactory concrete, Action<IParsable> onBefore, Action<IParsable> onAfter)
+        protected ParseNodeProxyFactory(IParseNodeFactory concrete, Action<IParsable> onBefore, Action<IParsable> onAfter)
```
calebkiage, line 30: Isn't this a breaking change?

baywet: No because the class is abstract so you can't instantiate it directly.

calebkiage (approved):

PR #345, opened 2024-08-27
fix/task management

baywet:
- **chore: adds task analyzers**
- **chore: linting of native wrapper**
- **fix: dispose async when possible**


andrueastman (approved):
👍🏼

Issue #347, opened 2024-08-30
InMemoryBackingStore performs poorly when reading properties


Le-Merle:
The BackingStore tries to keep the "dirty" state up to date on every call to read a property value. Because of the way the logic to track changes on collections was implemented, this has a detrimental impact on both CPU and memory consumption.

**Currently, reading a property from a model is an operation of potentially exponential complexity that allocates memory on the heap.**

Whether paying this cost on every property read in order to prevent sending additional data on the network is worth it or not is debatable, but it definitely isn't if we have no intention of performing any update call.

From my understanding of the implementation, the idea is to always return every values except when `ReturnOnlyChangedValues` is set to true. This seems to mostly be activated right before serialization, and deactivated immediately after.

If that is the case, I would suggest checking the value of `ReturnOnlyChangedValues` before calling `EnsureCollectionPropertyIsConsistent` when reading a property value. There is no rush to keep the dirty state synchronized until we are actually using it. On that note, it might also be a good idea to call it within `EnumerateKeysForValuesChangedToNull`, although I haven't exactly explored every intended usage of this function.

A local test with this modification applied produced large improvements on speed and removed any memory allocation when reading properties.

Is my understanding of the intent of InMemoryBackingStore correct? Would my suggested fix make sense?

baywet:
Hi @Le-Merle 
Thanks for using kiota and for reaching out.

The changes you're proposing sound coherent, I might be missing some details here. 

Can you please send a pull request so we can discuss the changes in details?

PR #368, opened 2024-09-03
Only track changed values when it matters

Le-Merle:
As requested in #347 

The idea is to skip tracking the change state until the `ReturnOnlyChangedValues` property is true.

### DIFF
--- a/src/abstractions/store/InMemoryBackingStore.cs
+++ b/src/abstractions/store/InMemoryBackingStore.cs
@@ -253,9 +253,9 @@ namespace Microsoft.Kiota.Abstractions.Store
 
         private void ForwardReturnOnlyChangedValuesToNestedInstances()
         {
-            foreach(var item in store)
+            foreach(var item in store.Values)
             {
-                if(item.Value.Item2 is Tuple<ICollection, int> collectionTuple)
+                if(item.Item2 is Tuple<ICollection, int> collectionTuple)
                 {
                     foreach(var collectionItem in collectionTuple.Item1)
                     {
@@ -263,7 +263,7 @@ namespace Microsoft.Kiota.Abstractions.Store
                         backedModel.BackingStore.ReturnOnlyChangedValues = _returnOnlyChangedValues;
                     }
                 }
-                else if(item.Value.Item2 is IBackedModel backedModel)
+                else if(item.Item2 is IBackedModel backedModel)
                 {
                     backedModel.BackingStore.ReturnOnlyChangedValues = _returnOnlyChangedValues;
                 }
--- a/src/abstractions/store/InMemoryBackingStore.cs
+++ b/src/abstractions/store/InMemoryBackingStore.cs
@@ -15,10 +15,21 @@ namespace Microsoft.Kiota.Abstractions.Store
     public class InMemoryBackingStore : IBackingStore
     {
         private bool isInitializationComplete = true;
+        private bool _returnOnlyChangedValues = false;
+
         /// <summary>
         /// Determines whether the backing store should only return changed values when queried.
         /// </summary>
-        public bool ReturnOnlyChangedValues { get; set; }
+        public bool ReturnOnlyChangedValues
+        {
+            get => _returnOnlyChangedValues;
+
+            set
+            {
+                _returnOnlyChangedValues = value;
+                ForwardReturnOnlyChangedValuesToNestedInstances();
+            }
+        }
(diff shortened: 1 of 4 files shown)


src/abstractions/store/InMemoryBackingStore.cs
```diff
@@ -128,6 +142,14 @@ public void Set<T>(string key, T? value)
         /// <returns>A collection of strings containing keys changed to null </returns>
         public IEnumerable<string> EnumerateKeysForValuesChangedToNull()
         {
+            if(ReturnOnlyChangedValues) // refresh the state of collection properties if they've changed in size.
+            {
+                foreach(var item in store)
```
baywet, line 147: we'd be enumerating the store twice, what do you think of moving the condition in the loop below instead?

src/abstractions/store/InMemoryBackingStore.cs
```diff
@@ -228,5 +250,24 @@ private void EnsureCollectionPropertyIsConsistent(string key, object? storeItem)
                 }
             }
         }
+
+        private void ForwardReturnOnlyChangedValuesToNestedInstances()
+        {
+            foreach(var item in store)
```
baywet, line 256: could we use store.Values to make the iteration a little more simple?

baywet (changes requested):
Thanks for the contribution!

Le-Merle:
@microsoft-github-policy-service agree company="Workleap"

src/abstractions/store/InMemoryBackingStore.cs
Le-Merle: It turns out there's a reason why `Enumerate()` uses 2 consecutive iterations.

Since `EnsureCollectionPropertyIsConsistent()` does not simply flip the dirty flag in-place but instead calls `.Set()`, which will replace the object in the store entirely, performing both operations in a single loop doesn't actually work as the store will be properly updated but the current iteration variable will stay the same. Therefore, if we use a single iteration, an item that should be detected as dirty will be properly updated in the store but will not be considered in the current enumeration as dirty.

A quick check gives me the feeling that the call to `.Set()` is done to properly support the event subscription, but I think it might be possible to change its implementation so it reuses/updates an existing value in the store rather than recreating a new one from scratch every time (this would probably improve memory allocation overall, too). Still, I think we're starting to stray from the goal of this PR here.

Do you want me to push the research further on how `.Set()` could be improved to support a single iteration here (and potentially save memory allocations when ensuring the collection consistency) or can we consider this PR as good enough?

baywet: > I don't mind moving the check in the same loop, but this pattern was based on the current implementation of `Enumerate()` right above it, which also enumerates the store twice.
> 
> Should I move both?

yes please, and enumerate could even yield return instead of building an intermediary list.

> Do you want me to push the research further on how .Set() could be improved to support a single iteration here (and potentially save memory allocations when ensuring the collection consistency) or can we consider this PR as good enough?

yes please, but timebox it. I don't want you to burn too much time on this.

Le-Merle: > yes please, but timebox it. I don't want you to burn too much time on this.
I've outran my timebox on this. For it to work, it basically requires a complete rewrite.

First the store itself needs to change, as it currently runs on immutable Tuples which are forcing to reassign a new object every time something changes, preventing a simple enumeration since the source is constantly modified. Switching to mutable objects with the intent to update them whenever possible means the `.Set()` function also needs to be rewritten, along with every check for Tuple/collections. At this point most of the class needs to be modified. This is a rather involved change and I'm not quite sure the level of effort/risk required is worth the gain - particularly in the scope of this PR.

> yes please, and enumerate could even yield return instead of building an intermediary list.
This could be a breaking change. The function is part of the interface, so anyone could currently be calling `Enumerate()` and changing the content of the store between yields, leading to undefined behavior. With the current implementation this works because of the intermediary list. I don't mind changing this, as long as we're aware of the consequences. On the positive side, it would allocate much less memory when iterating recursively within `EnsureCollectionPropertyIsConsistent()`.

baywet: thanks for the detailed information and for spending the time investigating this. Resolving for now.

baywet (dismissed):
I'd like @andrueastman go give it an additional review here.

andrueastman (approved):
👍🏼  on this change. 

Thanks for the contribution here @Le-Merle. Appreciate your time looking into this and making it better.

PR #417, opened 2024-10-02
Fix NuGet badges

martincostello:
Fix broken NuGet badges since `buildstats.info` was shuttered.


andrueastman (approved):
Thanks for the contribution here @martincostello

Issue #270, opened 2023-12-19
KiotaClientFactory doesn't build a httpclient that will work because it is missing auth middleware


darrelmiller:
https://github.com/microsoft/kiota-http-dotnet/blob/eb76ee893324bd373462e98dfdafeba95762feff/src/KiotaClientFactory.cs#L36

This issue is related to this ADR https://github.com/microsoft/kiota/issues/3941

We need to think about building an AuthHandler middleware that can use our Kiota Auth Providers so that we can have native HttpClient experiences.  Otherwise, we can't satisfy this ADR https://github.com/microsoft/kiota/issues/3200

baywet:
authored https://github.com/microsoftgraph/msgraph-sdk-design/pull/111 to help move this along

lvde0:
I stumbled upon this issue since I have a custom "Token Refresh" middleware that tries to re-authenticate the request with the refreshed token. I was wondering if it's possible to access the `IAuthenticationProvider` somehow to do that? I mean I can still manually tweak the authorization header, but I think it would be cleaner to go via `IAuthenticationProvider`.

baywet:
currently it's not, but with the new factory we'd be adding with this work, you could at least access it while initializing the middleware, and pass it to your custom middleware implementation.
We started the work in java https://github.com/microsoft/kiota-java/pull/1562

Issue #327, opened 2024-08-13
Add release-please configuration


Ndiritu:
We want:
 - [x] release please header file (yaml .github/release-please.yml)
 - [x] release please manifest (json .release-please-manifest.json)
 - [x] release please config (json release-please-config.json)
 - [x] changelog keep a changelog header and unreleased information to be removed
 - [x] contributing to be updated with commit formatting instructions
 - [x] delete any version bump scripts
 - [x] files where the version needs to be updated to be updated with the special comments for release please
 
 Sample config: https://github.com/microsoftgraph/msgraph-beta-sdk-java/pull/952

baywet:
(just made sure the RP app is installed on this repo)

baywet:
To do so for other repos:

1. JIT into the repo
1. In a separate tab, go to the [marketplace app](https://github.com/apps/release-please)
1. Click "configure"
1. Select Microsoft
1. Select the target repo in the drop down
1. click "update access"

PR #430, opened 2024-10-09
Adds Authorization handler

Ndiritu:
closes https://github.com/microsoft/kiota-dotnet/issues/270

### DIFF
--- a/tests/http/httpClient/Middleware/AuthorizationHandlerTests.cs
+++ b/tests/http/httpClient/Middleware/AuthorizationHandlerTests.cs
@@ -1,0 +1,114 @@
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using Microsoft.Kiota.Abstractions.Authentication;
+using Microsoft.Kiota.Http.HttpClientLibrary.Middleware;
+using Microsoft.Kiota.Http.HttpClientLibrary.Tests.Mocks;
+using Moq;
+using Xunit;
+
+namespace Microsoft.Kiota.Http.HttpClientLibrary.Tests.Middleware
+{
+    public class AuthorizationHandlerTests : IDisposable
+    {
+        private readonly MockRedirectHandler _testHttpMessageHandler;
+
+        private IAccessTokenProvider _mockAccessTokenProvider;
+
+        private readonly string _expectedAccessToken = "token";
+
+        private readonly string _expectedAccessTokenAfterCAE = "token2";
+        private AuthorizationHandler _authorizationHandler;
+        private readonly BaseBearerTokenAuthenticationProvider _authenticationProvider;
+        private readonly HttpMessageInvoker _invoker;
+
+        private readonly string _claimsChallengeHeaderValue = "authorization_uri=\"https://login.windows.net/common/oauth2/authorize\","
+                + "error=\"insufficient_claims\","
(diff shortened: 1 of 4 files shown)


tests/http/httpClient/Middleware/AuthorizationHandlerTests.cs
```diff
@@ -0,0 +1,117 @@
...
+using Microsoft.Kiota.Http.HttpClientLibrary.Middleware;
+using Microsoft.Kiota.Http.HttpClientLibrary.Tests.Mocks;
+using Moq;
+using Xunit;
+
+namespace Microsoft.Kiota.Http.HttpClientLibrary.Tests.Middleware
+{
+    public class AuthorizationHandlerTests : IDisposable
+    {
+        private readonly MockRedirectHandler _testHttpMessageHandler;
```
andrueastman, line 14: Since this is shared in the class, you may run into problems when the multiple tests use it as each instance is currently meant to only send one request. 

https://github.com/microsoft/kiota-dotnet/blob/75f6bd3ae9d4bff6ad0c9a1f38cbf5be7c7289d6/tests/http/httpClient/Mocks/MockRedirectHandler.cs#L18

You should probably use the `TestHttpMessageHandler` type instead

tests/http/httpClient/Middleware/AuthorizationHandlerTests.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        private IAccessTokenProvider _mockAccessTokenProvider;
+
+        private readonly string _expectedAccessToken = "token";
```
andrueastman, line 18: string properties can be `const` instead of readonly..

tests/http/httpClient/Middleware/AuthorizationHandlerTests.cs
Ndiritu: `MockRedirectHandler` allowed me to chain multiple requests easily since the Auth handler retries only once if a CAE challenge is returned by the first request: [here](https://github.com/microsoft/kiota-dotnet/pull/430/files#diff-d10ea300ecbd16c34c6980e04b9b192f660a99f284b061754539ce34c41d3448R122)

andrueastman: I would suggest removing this as a class variable and initializing a new varible in the tests. Since the tests are not guaranteed to be executed sequentially depending on test context, you may run into flaky tests...

```cs
var testHttpMessageHandler = new MockRedirectHandler();
```

andrueastman (commented):
We'll also need a version bump/changelog as we haven't closed https://github.com/microsoft/kiota-dotnet/issues/327 in this repo yet.

tests/http/httpClient/Middleware/AuthorizationHandlerTests.cs
Ndiritu: This would mean initialising a few more variables currently in the test class constructor within every test.
Curious why the tests might get flaky?
I thought each test will invoke the constructor & the class variables will be re-initialised, so each test uses a different instance of the handler. Similar approach is being used in the Redirect hander [tests](https://github.com/microsoft/kiota-dotnet/blob/main/tests/http/httpClient/Middleware/RedirectHandlerTests.cs)

andrueastman: Yeah. Your right. Resolving this for now.
https://xunit.net/docs/shared-context

andrueastman (approved):
👍🏼

Issue #210, opened 2024-03-21
Add function overloads with Type as parameter for KiotaSerializer.Deserialize<T>


dansmitt:
We (@MichaMican and I) have implemented an TextFormatter extension to use Kiota as thge standard serializer for `IParsable` objects when working with ASP.NET Controllers.

See: https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/2343#issuecomment-2012679960

In our TextInputFormatter we have to fallback to using Reflection because we work with an implementation of IParsable object that we have to set as the generic type of `KiotaSerializer.Deserialize<T>` during run time. As of now the function unfortunately has no overload to specify the Type as a parameter which would be super usefull in our case as we could then imrpove code readability by not using reflection.

Snippet from the TextInputFormatter:
```csharp
public override Task<InputFormatterResult> ReadRequestBodyAsync(InputFormatterContext context, Encoding encoding)
{
    var httpContext = context.HttpContext;

    if (httpContext.Request.Body.CanRead)
    {
        var targetType = context.ModelType; // <- This is where we get the Type for Deserialize per runtime

        #region Reflection to define Generic type T of KiotaJsonSerializer.Deserialize<T> dynamically
        var methodInfo = typeof(KiotaJsonSerializer).GetMethod("Deserialize", BindingFlags.Public | BindingFlags.Static, [typeof(Stream)]);
        if (methodInfo == null)
        {
            throw new Exception("KiotaJsonSerializer.Deserialize changed. Please update implementation");
        }

        // This line types the KiotaJsonSerializer.Deserialize<T> to T equals to targetType
        var methodWithForcedTargetType = methodInfo.MakeGenericMethod(targetType);

        // obj is null because KiotaJsonSerializer.Deserialize is a static function
        // This function is analog to KiotaJsonSerializer.Deserialize<T>(httpContext.Request.Body) -> Where T = targetType
        var deserializedObject = methodWithForcedTargetType.Invoke(null, [httpContext.Request.Body]);

       // As we used reflection deserializedObject is of type object, hence we have to continue to use Reflection to access the BackingStore/InitializationCompleted property.
        var backingStoreProperty = (deserializedObject?.GetType().GetProperty("BackingStore"));
        var backingStoreValue = backingStoreProperty?.GetValue(deserializedObject);

        backingStoreValue?.GetType().GetProperty("InitializationCompleted")?.SetValue(backingStoreValue, false);

        #endregion

[...]

PR #436, opened 2024-10-11
Added Non-Generic Solution to KiotaDeserialization

DerGuru:
This code addition fulfills the request of [Issue 210](https://github.com/microsoft/kiota-dotnet/issues/210), where this was explicitly requested.

@dansmitt is a co-worker of mine and I thought, that this might be of relevance to more than just him.

I used the same Unit-Tests, as you were, modifying them just as much, to use the newly introduced method signatures.


src/abstractions/serialization/KiotaJsonSerializer.Deserialization.Untyped.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Linq;
+
+#pragma warning disable IL3050 
+
+namespace Microsoft.Kiota.Abstractions.Serialization;
+
+public static partial class KiotaJsonSerializer
+{
+    private static bool IsIParsable(this Type? type) => type?.IsAssignableTo(typeof(IParsable)) ?? false;
+    private abstract class KiotaJsonDeserializer
```
baywet, line 24: can we give different names to this class and the one bellow to ease up code review?
It feels like this one should be called somethingFactory.
And the one at line 30 somethingAdapter.

Or something alone those lines.

src/abstractions/serialization/KiotaJsonSerializer.Deserialization.Untyped.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Threading.Tasks;
+using System.Threading;
+using System.IO;
+using System;
+using System.Collections.Concurrent;
+using System.Diagnostics.CodeAnalysis;
+using System.Collections.Generic;
+using System.Linq;
+
+#pragma warning disable IL3050
```
baywet, line 17: can you provide more context around this suppression please?

baywet (changes requested):
Thanks for the contribution!

src/abstractions/serialization/KiotaJsonSerializer.Deserialization.Untyped.cs
DerGuru: Hi baywet,

surely you could name them as you please, but to comment on your suggestions.
The factory is in fact an abstract base class having a factory method (well one line of it) and executes the actual Activator, which is then placed in line 30.

I created this much like in System.Security, where (f. e.) one calls the create method of the abstract AES, which returns an actual implementation of a derived class.

I tried to keep it as separate as possible and even the derived classes are privately nested classes of the abstract base.

If you want to call the class something different, you might call it a wrapper around the KitoJsonSerializer for this special case of having the target type as Type instead of handing it to a generic method. So I guess I will rename it.

I also got e broken build, which was the result of some wrong line endings and different whitespacing.
Is there something I could download for VS to have the style guide enforced before pushing it again to GitHub?

src/abstractions/serialization/KiotaJsonSerializer.Deserialization.Untyped.cs
DerGuru: Well the Warning I received was, that AOT will not have the type ready when using a native assembly and will have to create it during runtime. At least that is, what I got from the message and some reading on the internet.
 Nevertheless, this is intended behavior, because this is exactly for the case, when one does not know the Type at compile time and therefore needs this API in the first place. Otherwise one could have used the Generic Methods instead of these.

DerGuru:
Okay .. @baywet
I took your comments to heart and split the base class into a factory and an interface.
I also hope to have fixed the whitespace issue.... weirdly, that was auto generated code...

I was also able, to remove the `#pragma warning disable IL3050`, which just needed some attributes added to the code. And since I don't like to have pre compiler directives spread all over the code, I added the Attribute as a privately nested class, for .NET < 8. 
I hope, that is okay.

src/abstractions/serialization/KiotaJsonSerializer.Deserialization.NonGeneric.cs
```diff
@@ -0,0 +1,105 @@
...
+
+namespace Microsoft.Kiota.Abstractions.Serialization;
+public static partial class KiotaJsonSerializer
+{
+#if NET8_0_OR_GREATER
+#else
+    /// <summary>
+    /// Palceholder for .NET &lt; 8
+    /// </summary>
+    private class RequiresDynamicCodeAttribute : Attribute
```
baywet, line 24: Here is how we've solved the problem across the board.
https://github.com/microsoft/kiota-dotnet/blob/650812d9b598f192e6e55424e327d2bcee49c99a/src/abstractions/RequestInformation.cs#L51
Can you align with this please?

src/abstractions/serialization/KiotaJsonSerializer.Deserialization.NonGeneric.cs
```diff
@@ -0,0 +1,105 @@
...
+    /// <summary>
+    /// Deserializes the given stream into a collection of objects based on the content type.
+    /// </summary>
+    /// <param name="targetType">The target type to deserialize</param>
+    /// <param name="serializedRepresentation">The serialized representation of the object.</param>
+    /// <param name="cancellationToken">The cancellation token for the task</param>
+    [RequiresDynamicCode("Activator creates an instance of a generic class with the Target Type as the generic type argument.")]
+    public static Task<IEnumerable<IParsable>> DeserializeCollectionAsync(Type targetType, string serializedRepresentation, CancellationToken cancellationToken = default)
+         => KiotaJsonDeserializationWrapperFactory.Create(targetType).DeserializeCollectionAsync(serializedRepresentation, cancellationToken);
+}
```
baywet, line 104: why is this whole section duplicated instead of calling into the non media type specific one?

baywet (commented):
A couple additional questions/comments.

src/abstractions/serialization/KiotaJsonSerializer.Deserialization.NonGeneric.cs
DerGuru: done

DerGuru: Well it looks kind of ugly now ... but can't help that, I guess.

baywet: yes it does. That's ok, consistency over "prettyness". Once we address this, we'll be able to do so across the libraries.

baywet (dismissed):

I also took the liberty of making a bunch of "cleanup" changes here.

@andrueastman for final review and merge

andrueastman (commented):
Otherwise looks good from my end. Will update version/release notes on getting clarification on duplicate code question.

andrueastman (approved):

PR #493, opened 2024-12-07
feat: add BodyInspectionHandler

### LINKED ISSUE #482: http - implement body inspection handler
specification https://github.com/microsoftgraph/msgraph-sdk-design/pull/116
make sure you update the [documentation](https://learn.microsoft.com/en-us/openapi/kiota/middleware?tabs=csharp#middleware-handlers) as well

marcinjahn:
Closes #482

Docs updates done via https://github.com/MicrosoftDocs/openapi-docs/pull/140.

### DIFF
--- a/src/http/httpClient/Middleware/BodyInspectionHandler.cs
+++ b/src/http/httpClient/Middleware/BodyInspectionHandler.cs
@@ -60,12 +60,14 @@ public class BodyInspectionHandler : DelegatingHandler
         {
             if(options.InspectRequestBody)
             {
-                options.RequestBody = await CopyToStreamAsync(request.Content, cancellationToken).ConfigureAwait(false);
+                options.RequestBody = await CopyToStreamAsync(request.Content, cancellationToken)
+                    .ConfigureAwait(false);
             }
             var response = await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
             if(options.InspectResponseBody)
             {
-                options.ResponseBody = await CopyToStreamAsync(response.Content, cancellationToken).ConfigureAwait(false);
+                options.ResponseBody = await CopyToStreamAsync(response.Content, cancellationToken)
+                    .ConfigureAwait(false);
             }
 
             return response;
@@ -78,14 +80,12 @@ public class BodyInspectionHandler : DelegatingHandler
 #if NETSTANDARD2_1_OR_GREATER || NETCOREAPP3_0_OR_GREATER
         [return: NotNullIfNotNull(nameof(httpContent))]
 #endif
-        static async Task<Stream?> CopyToStreamAsync(HttpContent? httpContent, CancellationToken cancellationToken)
+        static async Task<Stream> CopyToStreamAsync(
+            HttpContent? httpContent,
+            CancellationToken cancellationToken
+        )
         {
-            if(httpContent == null)
-            {
-                return null;
-            }
-
-            if(httpContent.Headers.ContentLength == 0)
+            if(httpContent is null or { Headers.ContentLength: 0 })
             {
                 return Stream.Null;
             }
@@ -97,7 +97,6 @@ public class BodyInspectionHandler : DelegatingHandler
 #else
             await httpContent.CopyToAsync(stream).ConfigureAwait(false);
 #endif
-
 
             if(stream.CanSeek)
             {
--- a/src/http/httpClient/Middleware/BodyInspectionHandler.cs
+++ b/src/http/httpClient/Middleware/BodyInspectionHandler.cs
@@ -91,7 +91,13 @@ public class BodyInspectionHandler : DelegatingHandler
             }
 
             var stream = new MemoryStream();
+
+#if NET5_0_OR_GREATER
             await httpContent.CopyToAsync(stream, cancellationToken).ConfigureAwait(false);
+#else
+            await httpContent.CopyToAsync(stream).ConfigureAwait(false);
+#endif
+
 
             if(stream.CanSeek)
             {
--- a/src/http/httpClient/Middleware/BodyInspectionHandler.cs
+++ b/src/http/httpClient/Middleware/BodyInspectionHandler.cs
@@ -36,13 +36,13 @@ public class BodyInspectionHandler : DelegatingHandler
         CancellationToken cancellationToken
     )
     {
-        if (request == null)
+        if(request == null)
             throw new ArgumentNullException(nameof(request));
 
         var options = request.GetRequestOption<BodyInspectionHandlerOption>() ?? _defaultOptions;
 
         Activity? activity;
-        if (request.GetRequestOption<ObservabilityOptions>() is { } obsOptions)
+        if(request.GetRequestOption<ObservabilityOptions>() is { } obsOptions)
         {
             var activitySource = ActivitySourceRegistry.DefaultInstance.GetOrCreateActivitySource(
                 obsOptions.TracerInstrumentationName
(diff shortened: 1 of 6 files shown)


tests/http/httpClient/Middleware/BodyInspectionHandlerTests.cs
```diff
@@ -0,0 +1,112 @@
...
+        var request = new HttpRequestMessage(HttpMethod.Post, "https://localhost")
+        {
+            Content = new StringContent("request test")
+        };
+        var response = await invoker.SendAsync(request, default);
+
+        // Then
+        Assert.NotNull(option.RequestBody);
+        Assert.Equal("request test", GetStringFromStream(option.RequestBody!));
+    }
```
andrueastman, line 36: Since the test invoker does not read the http content(its not really sending it) should we add an assertion that the content is still intact/readable/valid?

src/http/httpClient/Middleware/BodyInspectionHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Threading.Tasks;
+using Microsoft.Kiota.Http.HttpClientLibrary.Extensions;
+using Microsoft.Kiota.Http.HttpClientLibrary.Middleware.Options;
+
+namespace Microsoft.Kiota.Http.HttpClientLibrary.Middleware;
+
+/// <summary>
+/// The Body Inspection Handler allows the developer to inspect the body of the request and response.
+/// </summary>
+public class BodyInspectionHandler : DelegatingHandler
```
andrueastman, line 20: Similar to the `HeadersInspectionHandler`, should we register this in the default list so that the option can be used without having to create a new client or modify/update the handler list? 
https://github.com/microsoft/kiota-dotnet/blob/216801b4931b03aac86aa9d656a472bf6379bee3/src/http/httpClient/KiotaClientFactory.cs#L91

src/http/httpClient/Middleware/BodyInspectionHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+        [return: NotNullIfNotNull(nameof(httpContent))]
+#endif
+        static async Task<Stream?> CopyToStreamAsync(HttpContent? httpContent)
+        {
+            if (httpContent == null)
+            {
+                return null;
+            }
+
+            var stream = new MemoryStream();
```
andrueastman, line 88: I suspect there are scenarios where the response/request stream may not be seekable(not buffered) so setting the position or reading the stream twice will throw. We should probably add code to handle that (skip the copy)

andrueastman (commented):
Thanks for the contribution here @marcinjahn 

Any chance you can also run `dotnet format` command from the project root to resolve the linting errors?

src/http/httpClient/Middleware/BodyInspectionHandler.cs
marcinjahn: done

tests/http/httpClient/Middleware/BodyInspectionHandlerTests.cs
marcinjahn: Makes sense, I added an assert, similar to how I did it for response assertion.

src/http/httpClient/Middleware/BodyInspectionHandler.cs
marcinjahn: I modified the `CopyToStreamAsync` method, have a look

marcinjahn:
> Thanks for the contribution here @marcinjahn
> 
> Any chance you can also run `dotnet format` command from the project root to resolve the linting errors?

I pushed `dotnet format`ted code now

src/http/httpClient/Middleware/BodyInspectionHandler.cs
```diff
@@ -0,0 +1,104 @@
...
+                return null;
+            }
+
+            if(httpContent.Headers.ContentLength == 0)
+            {
+                return Stream.Null;
+            }
+
+            var stream = new MemoryStream();
+            await httpContent.CopyToAsync(stream);
```
baywet, line 94: Add configure await ( couple of other places) 
We should be able to pass a cancellation token as well here

baywet (commented):
Thank you for the contribution!

src/http/httpClient/Middleware/BodyInspectionHandler.cs
andrueastman: Thanks. This makes sense. 
Do you think it would make sense to use this `TryComputeLength` instead?
https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpcontent.trycomputelength?view=net-8.0

src/http/httpClient/Middleware/BodyInspectionHandler.cs
marcinjahn: thanks, missed that. Should be alright now

src/http/httpClient/Middleware/BodyInspectionHandler.cs
```diff
@@ -0,0 +1,104 @@
...
+                return null;
+            }
+
+            if(httpContent.Headers.ContentLength == 0)
+            {
+                return Stream.Null;
+            }
+
+            var stream = new MemoryStream();
+            await httpContent.CopyToAsync(stream, cancellationToken).ConfigureAwait(false);
```
baywet, line 94: shouldn't we check whether the underlying streams are rewindable?
My concern here is the request body will be at its end position before it's sent over the network after the copy but before it's sent on the network.
Same thing the other way around with the response body and deserialization.

marcinjahn: When can that happen? How to check for such case?

marcinjahn: In unit tests, we're checking if  (after the copy) both request and response are still able to give us the proper content.

baywet: I believe string contents are buffered by default, which would explain why this issue does not appear with the current tests.
Using a memory stream as the source and a different kind of content (application octet stream for example) might reveal it.
Can you try to expose the issue by tweaking the tests please?

marcinjahn: will do

src/http/httpClient/Middleware/BodyInspectionHandler.cs
andrueastman: Sorry, I missed that. 
I believe we can go forward with the current one for now.

src/http/httpClient/Middleware/BodyInspectionHandler.cs
marcinjahn: @baywet Have a look at https://github.com/microsoft/kiota-dotnet/pull/493/commits/5ed13f77b7c1cb94cb9ab794ce92a0085f74a588

baywet: yes! thank you! resolving

baywet (dismissed):
Thanks for the contribution!

@andrueastman for final review and merge

marcinjahn:
@baywet @andrueastman Please have a look, I simplified it juat a little bit. The `RequestStream` and `ResponseStream` are no longer nullable, they're `Stream.Null` by default, and this is also the value being returned anytime there is no content in request/response. Previously, I'd return `null` or `Stream.Null` and probably it doens't make sense to do that, especially since it was not documented when each one gets returned.

baywet (approved):

baywet (approved):

PR #505, opened 2025-01-07
ci: setup release please

andrueastman:
Fixes https://github.com/microsoft/kiota-dotnet/issues/327


baywet (approved):

baywet:
@andrueastman interesting, they now add checks when you change the configuration, and it seems some JSON is invalid (see the checks panel)

andrueastman:
> and it seems some JSON is invalid (see the checks panel)

I'm not sure why it's failing here as the Json is valid.
I suspect the issue is that it's looking for the config files in main branch (which don't exist yet). We can probably try and merge and see if all is well?

baywet:
I tried removing the BOM from the encoding since it was complaining about index 0. I now do not see the checks anymore, I'm not sure whether this is because they are valid or because they don't trigger on push. Either way, we should be good to merge.

baywet (approved):

PR #519, opened 2025-02-05
feat: Add trim-safe handler lookup

### LINKED ISSUE #518: GetDefaultHandlerTypes causes trim warning for its intended usage.
`KiotaClientFactory.GetDefaultHandlerTypes()` gets exposed so consumers can iterate over it and add the types to a DI container per the documentation, but it currently causes a trim warning:

```cs
var types = KiotaClientFactory.GetDefaultHandlerTypes();
        
foreach (var type in types)
{
    collection.AddTransient(type);
}
```
`
IL2072: 'serviceType' argument does not satisfy 'DynamicallyAccessedMemberTypes. PublicConstructors' in call to 'Microsoft. Extensions. DependencyInjection. ServiceCollectionServiceExtensions. AddTransient(IServiceCollection, Type)'. The return value of method 'System. Collections. Generic. IEnumerator<T>.Current. get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
`

Per https://github.com/dotnet/runtime/issues/110239 the only two ways to fix it are to wrap the types in a struct or use a custom collection, both of which are breaking changes unfortunately.

Hawxy:
I went with the struct wrapper as it's a bit nicer than implementing a custom enumerator IMO.

Closes #518

### DIFF
--- a/src/http/httpClient/KiotaClientFactory.cs
+++ b/src/http/httpClient/KiotaClientFactory.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Net;
 using System.Net.Http;
 using Microsoft.Kiota.Abstractions;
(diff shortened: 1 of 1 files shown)


Hawxy:

baywet:
@Hawxy Thanks for the contribution!

As you can see the CI is failing because of formatting.
Can you please run `dotnet format` locally and push the changes?

Thanks!

Hawxy:
Has been sorted.

andrueastman (approved):

PR #527, opened 2025-03-05
Allow sending a full IBackedModel instance including all recursive properties

erikmav:
A common need for reusing objects from a GET to do a POST/PUT. See https://github.com/microsoft/kiota/issues/3883 .

### DIFF
--- a/src/abstractions/store/IBackingStore.cs
+++ b/src/abstractions/store/IBackingStore.cs
@@ -56,5 +56,11 @@ namespace Microsoft.Kiota.Abstractions.Store
         bool InitializationCompleted { get; set; }
         /// <value>Whether to return only values that have changed since the initialization of the object when calling the Get and Enumerate methods.</value>
         bool ReturnOnlyChangedValues { get; set; }
+        /// <summary>
+        /// Sets all fields recursively to "modified" so they will be sent in the next serialization.
+        /// This is useful to allow the model object to be reused to send to a POST or PUT call.
+        /// Do not use if you are using a sparse PATCH.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


erikmav:
This is a proposal implementation of the recommendation provided in https://github.com/microsoft/kiota/issues/3883 and taken from production code used in an Entra/MSGraph related service. Notably there are no parallel PRs (yet), I can provide those if this gets approved/committed.

src/abstractions/store/IBackingStore.cs
```diff
@@ -56,5 +56,11 @@ public interface IBackingStore
         bool InitializationCompleted { get; set; }
         /// <value>Whether to return only values that have changed since the initialization of the object when calling the Get and Enumerate methods.</value>
         bool ReturnOnlyChangedValues { get; set; }
+        /// <summary>
+        /// Sets all fields recursively to "modified" so they will be sent in the next serialization.
+        /// This is useful to allow the model object to be reused to send to a POST or PUT call.
+        /// Do not use if you are using a sparse PATCH.
+        /// </summary>
+        void MakeSendable();
```
andrueastman, line 64: Adding this to a public interface would be breaking change. 
Maybe we should look into a solution similar to https://github.com/microsoft/kiota-dotnet/commit/9c33bbde0d09cd5a8fa32220414ab9444d7724c0 where we update the serialization registry to return a serializer that will ignore the backing store.

erikmav: Examining alternate approaches. Microsot.Graph is not up to date with that commit, awaiting that to be completed.

Issue #529, opened 2025-03-12
MultiPartBody: Impossible to add multiple Body Parts with the same name


romanett:
### What are you generating using Kiota, clients or plugins?

API Client/SDK

### In what context or format are you using Kiota?

Nuget tool

### Client library/SDK language

Csharp

### Describe the bug

I am using an existing API with the Atteached API Spec.

It expects a multipart body with an array of files:

e.g swagger Editor lets you paste multiple files:

![Image](https://github.com/user-attachments/assets/9f1bbf64-d996-42bb-8d99-fbdd5d518416)

However the KIOTA Multipart body uses a dictionary based on the Part name, so this does not allow you to create such a body:

```C#
            body.AddOrReplacePart<byte[]>("files", "text/plain", file, "test.txt");
            body.AddOrReplacePart<byte[]>("files", "text/plain", file, "test2.txt");

 if(!_parts.TryAdd(partName, value))
 {
     _parts[partName] = value;
 }
```
This creates such a request:

RequestRaw: 

```
--d11b4f38539b40fbb6a300dfddcede33
Content-Type: text/plain
Content-Disposition: form-data; name="files"; filename="test.txt"

Test

--d11b4f38539b40fbb6a300dfddcede33--
```

However I want to create such an request:

```
--d11b4f38539b40fbb6a300dfddcede33
Content-Type: text/plain
Content-Disposition: form-data; name="files"; filename="test.txt"

Test

--d11b4f38539b40fbb6a300dfddcede33
Content-Type: text/plain
Content-Disposition: form-data; name="files"; filename="test2.txt"

Test2

--d11b4f38539b40fbb6a300dfddcede33--
```

### Expected behavior

Be able to create a multipart body with multiple elements with the same name.

### How to reproduce

Generate client with attached API spec.

Try to add multiple files

### Open API description file

```

openapi: 3.0.1
info:
  title: Test
  version: 1.0.0.0
paths:
  /v6/mailbox/messages:
    put:
      summary: Insert a message into a Mailbox
      description: |
        ## This endpoint inserts a message into the specified mailbox.
        **Returns** an object with ids for the newly created message
      operationId: putMessageV6
      requestBody:

        content:
          multipart/form-data:
            schema:
              $ref: '#/components/schemas/CreateMessage'
            encoding:
              files:
                contentType: application/octet-stream
      responses:
        '200':
          description: |
            The id of the created message is returned.
components:
  schemas:
    CreateMessage:
      type: object
[...]

PR #530, opened 2025-03-20
MultiPartBody: Add Support for multiple Body Parts with the same name

romanett:
- change MultiPartbody internal dictionary key from string to Tuple<string, string?)
- Add additional method overloads for GetPartValue<T> RemovePart
- Add EqualityComparer fro Tuple<string, string?>
- Add failing Test showcasing the imminent breaking change
- Add passing Test verifying the new intened behaviour

Fixes: #529

### DIFF
--- a/src/abstractions/MultipartBody.cs
+++ b/src/abstractions/MultipartBody.cs
@@ -63,7 +63,21 @@ public class MultipartBody : IParsable
     /// <returns>The value of the part.</returns>
     public T? GetPartValue<T>(string partName)
     {
-        return GetPartValue<T>(partName, null);
+        var value = GetPartValue<T>(partName, null);
+
+        if(EqualityComparer<T?>.Default.Equals(value, default))
+        {
+            foreach(var key in _parts.Keys)
+            {
+                if(key.Item1 == partName)
+                {
+                    value = GetPartValue<T>(partName, key.Item2);
+                    break;
+                }
+            }
+        }
+
+        return value;
     }
     /// <summary>
     /// Gets the value of a part from the multipart body.
@@ -95,7 +109,21 @@ public class MultipartBody : IParsable
     /// <returns>True if the part was removed, false otherwise.</returns>   
     public bool RemovePart(string partName)
     {
-        return RemovePart(partName, null);
+        bool success = RemovePart(partName, null);
+
+        if(!success)
+        {
+            foreach(var key in _parts.Keys)
+            {
+                if(key.Item1 == partName)
+                {
+                    success = RemovePart(partName, key.Item2);
+                    break;
+                }
+            }
+        }
+
+        return success;
     }
 
     /// <summary>
--- a/src/abstractions/MultipartBody.cs
+++ b/src/abstractions/MultipartBody.cs
@@ -47,10 +47,11 @@ public class MultipartBody : IParsable
         {
             throw new ArgumentNullException(nameof(partValue));
         }
+        var key = Tuple.Create(partName, fileName);
         var value = new Part(partName, partValue, contentType, fileName);
-        if(!_parts.TryAdd(partName, value))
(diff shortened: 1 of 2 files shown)


romanett:

src/abstractions/MultipartBody.cs
```diff
@@ -47,10 +47,11 @@ public void AddOrReplacePart<T>(string partName, string contentType, T partValue
         {
             throw new ArgumentNullException(nameof(partValue));
         }
+        var key = Tuple.Create(partName, fileName);
```
baywet, line 50: do you think we could use value tuples instead?
It's a small change but should reduce allocations.
https://learn.microsoft.com/en-us/dotnet/api/system.valuetuple?view=net-9.0

baywet (changes requested):
Thanks for the contribution!

src/abstractions/MultipartBody.cs
```diff
@@ -60,12 +61,23 @@ public void AddOrReplacePart<T>(string partName, string contentType, T partValue
...
+    /// <param name="fileName">An optional file name for the part.</param>
+    /// <returns>The value of the part.</returns>
+    public T? GetPartValue<T>(string partName, string? fileName)
     {
         if(string.IsNullOrEmpty(partName))
         {
             throw new ArgumentNullException(nameof(partName));
         }
-        if(_parts.TryGetValue(partName, out var value))
+        if(_parts.TryGetValue(Tuple.Create(partName, fileName), out var value))
```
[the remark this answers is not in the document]
romanett: @baywet still neede for value tuple?

What shall we do about Test case: ImplementationBreaksExisting?

Accept this change or make the new / old behaviour configurable?

src/abstractions/MultipartBody.cs
```diff
@@ -196,4 +221,20 @@ private sealed class Part(string name, object content, string contentType, strin
...
+        public bool Equals((string, string) x, (string, string) y)
+        {
+            return StringComparer.OrdinalIgnoreCase.Equals(x.Item1, y.Item1) &&
+                   StringComparer.OrdinalIgnoreCase.Equals(x.Item2, y.Item2);
+        }
+
+        public int GetHashCode(ValueTuple<string, string?> obj)
+        {
+            int hash1 = StringComparer.OrdinalIgnoreCase.GetHashCode(obj.Item1);
+            int hash2 = obj.Item2 != null ? StringComparer.OrdinalIgnoreCase.GetHashCode(obj.Item2) : 0;
```
baywet, line 236:
```suggestion
            int hash1 = StringComparer.Ordinal.GetHashCode(obj.Item1);
            int hash2 = obj.Item2 != null ? StringComparer.Ordinal.GetHashCode(obj.Item2) : 0;
```

tests/abstractions/MultipartBodyTests.cs
```diff
@@ -90,4 +90,37 @@ public void WorksWithoutFilename()
...
+
+    [Fact]
+    public void ImplementationBreaksExisting()
+    {
+        var body = new MultipartBody();
+
+        body.AddOrReplacePart("file", "application/json", "fileContent", "file.json");
+
+        // existing usecase, file should still be able to be retreived
+        Assert.Equal("fileContent", body.GetPartValue<string>("file"));
```
baywet, line 124: could we implement a fallback for compatibility?
e.g. (for removal, get or update), if it's a call the the method without the file name parameter, and the subcall to the method with the parameter did NOT return anything, loop on the keys until we find a matching one that has the part name, and call again the sub method with the right key this time.

baywet (approved):

@musale for an additional review here

baywet (approved):

romanett:
@baywet do you have a rough timeline when the next release will happen, that includes this fix?

musale (approved):
thank you for the good work @romanett!

Issue #550, opened 2025-06-16
Handling "numbers as strings" when dealing with misbehaving APIs


earloc:
### What are you generating using Kiota, clients or plugins?

API Client/SDK

### In what context or format are you using Kiota?

Nuget tool

### Client library/SDK language

Csharp

### Describe the bug

We`re consuming an API, which specifies certain response-properties as "numbers" in it's openapi, but spits out those numbers "as strings" at runtime.

With manual response-parsing using f.e. System.Text.Json, this is a no-brainer these days most of the time (by leveraging [JsonSerializerOptions.NumberHandling](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.numberhandling?view=net-9.0)).

However, to enable this with Kiota generated clients, we need to do quite some gymnastics to get this in a similar state.

So far I only managed to get this running, by basically copy-pasting together a custom [JsonParseNodeFactory](https://github.com/microsoft/kiota-serialization-json-dotnet/blob/main/src/JsonParseNodeFactory.cs) along with an accompanying custom [JsonParseNode](https://github.com/microsoft/kiota-serialization-json-dotnet/blob/main/src/JsonParseNode.cs).

(Will provide a minimal reproducable sample in a comment).

### Expected behavior

When constructing a generated ApiClient, passing in a customized [KiotaJsonSerializationContext](https://github.com/microsoft/kiota-serialization-json-dotnet/blob/main/src/KiotaJsonSerializationContext.cs) should enable support for parsing numbers as strings out of the box. 

f.e.:

```csharp

var context = new KiotaJsonSerializationContext(new JsonSerializerOptions(JsonSerializerOptions.Web)); // can handle numbers-as-strings per default

        var apiClient = new ApiClient(new HttpClientRequestAdapter
        (
            authenticationProvider: new AnonymousAuthenticationProvider(),
            parseNodeFactory: new JsonParseNodeFactory(context),
            serializationWriterFactory: new JsonSerializationWriterFactory(context),
            httpClient: client
        ));
```

[...]

PR #551, opened 2025-06-17
Feat: deserialize numbers as strings

earloc:
# Synopsis

Enables opt-in support for deserializing "numbers as strings", by providing a configured `KiotaJsonSerializationContext`, in accordance to [JsonSerializerOptions.NumberHandling](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.numberhandling?view=net-9.0).

This is to provide a more streamlined approach when dealing with APIs, which may return numbers as strings, even when their openAPI specifies a more specific type, like 

```json
"type": "number",
"format": "double"
```

where the current behavior is to fail silently, while populating such properties with `null`.

Closes https://github.com/microsoft/kiota-dotnet/issues/550

## via `JsonSerializerOptions.Web`
```csharp
var context = new KiotaJsonSerializationContext(new JsonSerializerOptions(JsonSerializerOptions.Web)); // can handle numbers-as-strings per default

var apiClient = new ApiClient(new HttpClientRequestAdapter
(
    parseNodeFactory: new JsonParseNodeFactory(context),
    httpClient: client
));
```

## via customized `JsonSerializerOptions`
```csharp
var context = new KiotaJsonSerializationContext(new JsonSerializerOptions()
{
    NumberHandling = JsonNumberHandling.AllowReadingFromString
}
var apiClient = new ApiClient(new HttpClientRequestAdapter
(
    parseNodeFactory: new JsonParseNodeFactory(context),
    httpClient: client
));
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,6 +53,6 @@ The recommended commit types used are:
 - __docs__ for documentation updates e.g. ReadMe update or code documentation updates
 - __build__ for build system changes (gradle updates, external dependency updates)
 - __ci__ for CI configuration file changes e.g. updating a pipeline
-- __chore__ for miscallaneous non-sdk changesin the repo e.g. removing an unused file
+- __chore__ for miscallaneous non-sdk changes in the repo e.g. removing an unused file
 
 Adding an exclamation mark after the commit type (`feat!`) or footer with the prefix __BREAKING CHANGE:__ will cause an increment of the _major_ version.
--- a/src/serialization/json/JsonParseNode.cs
+++ b/src/serialization/json/JsonParseNode.cs
@@ -113,7 +113,7 @@ namespace Microsoft.Kiota.Serialization.Json
         /// Get the double value from the json node
         /// </summary>
         /// <returns>A double value</returns>
-        public double? GetDoubleValue() => _jsonNode.ValueKind == JsonValueKind.Number
+        public double? GetDoubleValue() => (_jsonNode.ValueKind == JsonValueKind.Number || (_jsonNode.ValueKind == JsonValueKind.String && _jsonSerializerContext.Options.NumberHandling.HasFlag(JsonNumberHandling.AllowReadingFromString)))
             ? _jsonNode.Deserialize(_jsonSerializerContext.Double)
             : null;
 
--- a/src/serialization/json/JsonParseNode.cs
+++ b/src/serialization/json/JsonParseNode.cs
@@ -97,7 +97,7 @@ namespace Microsoft.Kiota.Serialization.Json
         /// Get the float value from the json node
         /// </summary>
         /// <returns>A float value</returns>
-        public float? GetFloatValue() => _jsonNode.ValueKind == JsonValueKind.Number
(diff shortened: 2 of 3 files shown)


baywet (dismissed):
Thanks for the contribution!

baywet:
@earloc thanks for the very detailed analysis here in addition to this great pull request!
I think that as a general design principal, we're not afraid of throwing when things go completely off the rails, instead of silently failing, which can be much harder to investigate for people running into the scenario.

I believe this means the tests need to be updated with a couple of assert.throw for this case + formatting needs to be fixed.

earloc:
@baywet okidokey, I refactored the tests to reflect this behavior (throwing, when opted-in for reading "numbers-as-strings" and encountering a non-deserializable value.

Also ran a `dotnet format`... anything else regarding formatting

baywet (approved):

baywet:
Thank you for the additional information.

Yes, the information on this is sparse...

Just to confirm, the exceptions are ONLY going to be thrown on netfx when the JsonNumberHandling.AllowReadingFromString is used but no additional converter is passed?
It's not going to throw when no options are passed (vast majority of cases)
It's not going to throw when an option is passed, and while running on net5+

If that's an accurate depiction of the situation, I believe adding a documentation comment (remark or equivalent) for the Kiota serialization context to detail that scenario should be sufficient.
Please who really want the number as string parsing to work can bring additional converters by deriving from that type. If this is a common scenario, we can think of adding those with conditional compilation as well.

What do you think?

earloc:
@baywet  yes, wonderful wrapup of the current situation. Here's a differrent take to showcase this:

| runtime | AllowReadingFromString | runtime value | result | behavior | comment |
|--------|--------|--------|--------|-----|----|
| > net5 | no | `42` | `42` | unchanged | |
| > net5 | no | `"42"` | `null` | unchanged ||
| > net5 | no | `"not-a-number"` | `null` | unchanged ||
| > net5 | yes | `42` | `42` | unchanged ||
| > net5 | yes | `"42"` | `42` | new ||
| > net5 | yes | `"not-a-number"` | `throws` | new ||
| netfx | no | `42` | `42` | unchanged ||
| netfx | no | `"42"` | `null` | unchanged ||
| netfx | no | `"not-a-number"` | `null` | unchanged ||
| netfx | yes | `42` | `42` | unchanged ||
| netfx | yes | `"42"` | `throws` | new | different from >net5, needs additional JsonConverter to work|
| netfx | yes | `"not-a-number"` | `throws` | new ||


In general, we could bring this w/o any differrence in runtime-behavior for `netfx`, if we just conditionally compile the underlying `Get*Value`-methods to behave as before this change on any `< net5` fx.

But I rather chose to accept this behavior (and cover it trhroug the testcases) and NOT do conditional compilation for this, as it:
- adds complexitiy to the code-base
- may help someone still stuck on `netfx` to deal with "numbers-as-strings", even if there might be additional work needed (by adding respective `JsonConverters`)


I pushed again, hopefully letting CI pass (worked on my machine ;) )

earloc:
If you might be so kind and point my nose to where this could be documented, I might jump in there as well.
And again thx for the tight feedback loop 👍.

baywet (approved):

PR #566, opened 2025-09-19
fix: remove hardcoded HttpRequestMessage.Version from HttpClientRequestAdapter

### LINKED ISSUE #565: Client HttpRequestMessage.Version is hardcoded to 2.0 making it incompatible with IIS & Windows authentication
I am unable to use the current client in combination with IIS and Windows authentication.

I set up the HttpClient for DI like so:

```C#
serviceCollection
  .AddHttpClient<MyClientFactory>()
  .ConfigurePrimaryHttpMessageHandler(static _ => new SocketsHttpHandler
  {
      Credentials = CredentialCache.DefaultCredentials,
      PreAuthenticate = true
  })
```

And use the `AnonymousAuthenticationProvider` as `IAuthenticationProvider`.

This should be all that is needed to get the service's Windows credentials to be forwarded to the IIS website configured for Windows authentication.

However, in [HttpClientRequestAdapter.GetRequestMessageFromRequestInformation](https://github.com/microsoft/kiota-dotnet/blob/main/src/http/httpClient/HttpClientRequestAdapter.cs#L594C1) the `HttpRequestMessage.Version` is hardcoded to HTTP 2.0.

This is a problem as can be seen here [HTTP/2 on IIS, When is HTTP/2 Not Supported?](https://learn.microsoft.com/en-us/iis/get-started/whats-new-in-iis-10/http2-on-iis#when-is-http2-not-supported):
`Windows authentication (NTLM/Kerberos/Negotiate) is not supported with HTTP/2. In this case IIS will fall back to HTTP/1.1.`

The IIS logs show the credentials to be missing and IIS responds with a HTTP 401.2 (0x80070005: Access is denied).

When I comment the line `Version = new Version(2, 0)` it works.

What is the reason for hardcoding a version?
Can explicitly setting the version be removed? Note that the version can already be set on the `SocketsHttpHandler` level.

Rans4ckeR:
Closes #565

### DIFF
--- a/src/http/httpClient/KiotaClientFactory.cs
+++ b/src/http/httpClient/KiotaClientFactory.cs
@@ -4,7 +4,9 @@
 
 using System;
 using System.Collections.Generic;
+#if NET5_0_OR_GREATER
 using System.Diagnostics.CodeAnalysis;
+#endif
 using System.Net;
 using System.Net.Http;
 using Microsoft.Kiota.Abstractions;
@@ -26,20 +28,7 @@ namespace Microsoft.Kiota.Http.HttpClientLibrary
         /// <param name="optionsForHandlers">A array of <see cref="IRequestOption"/> objects passed to the default handlers.</param>
         /// <returns>The <see cref="HttpClient"/> with the default middlewares.</returns>
         public static HttpClient Create(HttpMessageHandler? finalHandler = null, IRequestOption[]? optionsForHandlers = null)
-        {
-            var defaultHandlersEnumerable = CreateDefaultHandlers(optionsForHandlers);
-            int count = 0;
-            foreach(var _ in defaultHandlersEnumerable) count++;
-
-            var defaultHandlersArray = new DelegatingHandler[count];
-            int index = 0;
-            foreach(var handler2 in defaultHandlersEnumerable)
-            {
-                defaultHandlersArray[index++] = handler2;
-            }
-            var handler = ChainHandlersCollectionAndGetFirstLink(finalHandler ?? GetDefaultHttpMessageHandler(), defaultHandlersArray);
-            return handler != null ? new HttpClient(handler) : new HttpClient();
-        }
+            => Create(finalHandler, CreateDefaultHandlers(optionsForHandlers));
 
         /// <summary>
         /// Initializes the <see cref="HttpClient"/> with a custom middleware pipeline.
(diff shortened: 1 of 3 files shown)


baywet:
Thank you for the additional information.

In an ideal world, we wouldn't even have to worry about it. The default version the client would try connections with would be determined by the host/runtime. And then on first connection, if a given version is not supported by the service, older versions would be tried, and the uppermost supported version would be "cached" during the lifetime of the client for the given service.

This is what OkHttp does for Java for example. And this is what HttpClient does for SSL/TLS/... versions. So I'm not sure why the Http version of the Http client does not also do that...

Short of that, the client application should have control, and we should have a default, instead of hardcoding things and locking people out in some edge cases. Please go ahead and update this pull request to the less elegant solution :)

Rans4ckeR:
I had some trouble with the obsolete `Microsoft.DotNet.PackageValidation` so I swapped it out with [Package validation](https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/package-validation/overview). I think this can be safely reverted however if needed.

The current solution needs some feedback.

For NET5_0_OR_GREATER:
- Always use the `HttpClientRequestAdapter` ctr parameters for `HttpVersion` & `HttpVersionPolicy` when provided.
- Otherwise, use the injected `HttpClient` `HttpVersion` & `HttpVersionPolicy`. This is the scenario I personally would always use. The provided `HttpClient` is configured by the client application in exactly the way it wants.

For NETSTANDARD2_1_OR_GREATER:
- Always use the `HttpClientRequestAdapter` ctr parameter for `HttpVersion` when provided.
- Otherwise always use `HttpVersion.Version20`.

For NETSTANDARD2_0_OR_GREATER || NETFRAMEWORK:
- Always use the `HttpClientRequestAdapter` ctr parameter for `HttpVersion` when provided.
- Otherwise always use `HttpVersion.Version11`.

Remarks:
- netstandard2.0 and net462 were also using HTTP2 until now, is HTTP2 supported however by these? Now it will default to HTTP11.
- Tests for netstandard2.0 and netstandard2.1 will not be run since the test project doesn't target these. xunit does not support this.
- Does the MsGraph SDK currently also configure its `HttpClient` to use HTTP2? If not it might have to be changed explicitely with the current solution. On the plus side they could also switch to HTTP3 without needing a change in kiota.

src/http/httpClient/HttpClientRequestAdapter.cs
```diff
@@ -54,6 +82,13 @@ public HttpClientRequestAdapter(IAuthenticationProvider authenticationProvider,
             sWriterFactory = serializationWriterFactory ?? SerializationWriterFactoryRegistry.DefaultInstance;
             obsOptions = observabilityOptions ?? new ObservabilityOptions();
             activitySource = ActivitySourceRegistry.DefaultInstance.GetOrCreateActivitySource(obsOptions.TracerInstrumentationName);
+
+#if NET5_0_OR_GREATER
+            this.httpVersion = httpVersion ?? client.DefaultRequestVersion;
+            this.httpVersionPolicy = httpVersionPolicy ?? client.DefaultVersionPolicy;
```
baywet, line 88: if we can get that information from the client in net5+, why do we even need to have a parameter/fields? Can't we simply read it directly from the client that's being passed? and if the consumer wants to tweak things, they can simply create the http client and set it to whatever they need.

baywet (commented):

Great work on the package validation change.

I left a few comments to help get it to its final stage.

For net5+, I think we're on the same page and additional changes should be made.

For graph defaults, [it's effectively using the same as kiota](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/67a8611ea39c0be68ee7db7371c15ec13cecb018/src/Microsoft.Graph.Core/Requests/GraphClientFactory.cs#L173)

baywet:
Thanks for looking into this in details.

> Is it an option to pair this change with a change in the MsGraph SDK pinning the Version to HTTP2 there? I feel it makes more sense to have the version pinned there for MsGraph than to pin the version in kiota affecting all kiota consumers.

Yes, this is something that's been done historically, we'll need to ensure the change there also includes bumping the dependency from kiota to ensure things are aligned.
I have not retained approval permissions on the Microsoft Graph repos though, so I'll have to coordinate with someone else internally.

src/http/httpClient/KiotaClientFactory.cs
```diff
@@ -26,20 +28,7 @@ public static class KiotaClientFactory
...
-            var defaultHandlersEnumerable = CreateDefaultHandlers(optionsForHandlers);
-            int count = 0;
-            foreach(var _ in defaultHandlersEnumerable) count++;
-
-            var defaultHandlersArray = new DelegatingHandler[count];
-            int index = 0;
-            foreach(var handler2 in defaultHandlersEnumerable)
-            {
-                defaultHandlersArray[index++] = handler2;
-            }
```
Rans4ckeR, line 39: This does a conversion from IList to an array. I replaced it with a collection initializer.
The same applies to the other Create() overload.

PR #705, opened 2026-04-22
Set CLSCompliant attribute where possible

WolfgangHG:
We set "CLSCompliant" it to "true" in our projects because our application provides a VB.NET based scripting, and thus all libraries that could be accessed by the VB.NET code should be CLSCompliant (e.g. to avoid the situation "protected member variable and property for this variable with same name, only differing in casing").
After having added the first Kiota generated REST client, I had to disable the "not CLSCompliant" warnings for this client with a ".editorconfig" file.

This pull request tries to make Kiota CLSCompliant.

It is not possible for the project "Microsoft.Kiota.Authentication.Azure" because it has public api using classes from Azure.Core that is also not CLSCompliant. That's why I set the attribute per project.

Only problem:
`IParseNode` defines a propery of type "sbyte" that is not CLSCompliant:

https://github.com/microsoft/kiota-dotnet/blob/9e4d0e3c8a783734f0d766a20b021d582cd7ee24/src/abstractions/serialization/IParseNode.cs#L39-L43

To my understanding, I can make the projects CLSCompliant anyway, and I just have to disable the warning for all code parts that are not CLS compliant.
This is done by setting `[CLSCompliant(false)]` where possible. Only the `IParseNode` interface requires a `#pragma warning disable CS3002` because the attribute is not allowed on interface members.

In project "Microsoft.Kiota.Serialization.Json", there was a warning `\obj\Debug\netstandard2.1\System.Text.Json.SourceGeneration\System.Text.Json.SourceGeneration.JsonSourceGenerator\MySerialKiotaJsonSerializationContext.SByte.g.cs(19,84,19,89): warning CS3003: Type of 'KiotaJsonSerializationContext.SByte' is not CLS-compliant`. that was not suppressible with a ".editconfig" file - it does not accept paths in the "obj" directory.
I first created a roslyn issue (https://www.github.com/dotnet/roslyn/issues/83232), and the bot pointed me to a workaround (https://www.github.com/dotnet/roslyn/issues/47384#issuecomment-3529305512): add a ".globalconfig" file in the project that disables the warning:

```
dotnet_diagnostic.CS3003.severity = silent
```
This affects also generated files.

Then add a ".editorconfig" file that enabled the warning again:

```
[*.{cs}]
dotnet_diagnostic.CS3003.severity = warning
```

This file does NOT affect generated files, so the warning from them is still suppressed by ".globalconfig".

As there was also a suppression workaorund for "CS1591: Missing XML comment..." in "Microsoft.Kiota.Serialization.Json" that should probably also affect only generated files, I moved this suppression also to the new files.

Roslyn already contains a fix (https://www.github.com/dotnet/roslyn/pull/82224) so that you can specify suppressions for generated paths in ".editorconfig", but this fix does not seem to be released.

Hope my idea is acceptable ;-)


src/abstractions/serialization/IParseNode.cs
```diff
@@ -40,7 +40,9 @@ public interface IParseNode
         ///  Gets the sbyte value of the node.
         /// </summary>
         /// <returns>The sbyte value of the node.</returns>
+#pragma warning disable CS3002 // Return type is not CLS-compliant
```
baywet, line 43: if we suppress this here, what's going to be the impact for people using the interface to access the method? (instead of the implementation type), are they going to get warnings?

src/serialization/json/.editorconfig
```diff
@@ -0,0 +1,15 @@
+#Part 2 of a workaround for two types of warning in files generated by the JsonSerializer:
+#"kiota-dotnet\src\serialization\json\obj\Debug\netstandard2.1\System.Text.Json.SourceGeneration\System.Text.Json.SourceGeneration.JsonSourceGenerator\KiotaJsonSerializationContext.SByte.g.cs(12,99,12,104): error CS3003: Type of 'KiotaJsonSerializationContext.SByte' is not CLS-compliant"
+#Those warnings cannot be disabled by a path filter in a ".editorconfig" file.
+#See https://github.com/dotnet/roslyn/issues/83232: disable the warnings in ".globalconfig".
+#Then re-enable them with a ".editorconfig" file (which affects only non-generated files)
+#
+#Suppressing the warning in ".editconfig" only might work in the future if the fix https://github.com/dotnet/roslyn/pull/82224 is released.
+#It will require a special path expression like "[$generated$/**/*.cs]"
+
+[*.{cs}]
```
baywet, line 10: can we make the filter more specific so we catch regressions?

src/serialization/json/.globalconfig
```diff
@@ -0,0 +1,10 @@
+#Part 1 of a workaround for two types of warning in files generated by the JsonSerializer:
+#"kiota-dotnet\src\serialization\json\obj\Debug\netstandard2.1\System.Text.Json.SourceGeneration\System.Text.Json.SourceGeneration.JsonSourceGenerator\KiotaJsonSerializationContext.SByte.g.cs(12,99,12,104): error CS3003: Type of 'KiotaJsonSerializationContext.SByte' is not CLS-compliant"
+#Those warnings cannot be disabled by a path filter in a ".editorconfig" file.
+#See https://github.com/dotnet/roslyn/issues/83232: disable the warnings in ".globalconfig".
+#Then re-enable them with a ".editorconfig" file (which affects only non-generated files).
+
+#CS1591: Missing XML comment for publicly visible type or member 'KiotaJsonSerializationContext.Boolean'
+dotnet_diagnostic.CS1591.severity = silent
```
baywet, line 8: aren't those neutralizing for the whole project effectively? can we get something more specific in place to prevent future regressions?

baywet (commented):
Thanks for the contribution!

src/abstractions/serialization/IParseNode.cs
WolfgangHG: @bawyet Well, this snippet in another C# project (which also was CLSCompliant) using the CLSCompliant version of "Microsoft.Kiota.Abstractions" does not print a warning:

```c#
public void DoSomething()
{
  IParseNode parseNode = null;
  SByte? sByte = parseNode.GetSbyteValue();
}
```

Using an interface implementation does not print a warning either:
```c#
public void DoSomething()
{
  JsonParseNode parseNode = new JsonParseNode(new JsonElement());
  SByte? sByte = parseNode?.GetSbyteValue();
}
```

But "no warning" is probably OK, as I tested it in a C# and a VB.NET library which both support the data type "SByte".

The interface method `GetSbyteValue` will probably be used only in Kiota generated code where also a public property will be generated - and this one causes a "CLSCompliant" warning if the generated client project is declared as CLSCompliant.

It would probably cause a compilation error if the target language does not support "SByte" anyway. Do you know of any .NET based language that does not support non-CLS-compliant types like "UInt32"?

And as I asked myself where this "SByte" is coming from: it is the [JsonSchema](https://ubjson.org/type-reference/value-types/#numeric) type `int8`.

What do you think? The `SByte` problem might cancel this whole pull request.

src/serialization/json/.globalconfig
WolfgangHG: @baywet No, the ".globalconfig" file is only used for the "Microsoft.Kiota.Serialization.Json" project, the other projects are not affected.

src/serialization/json/.editorconfig
WolfgangHG: Sorry, what do you mean ;-)?

The filter affects all .cs files that are part of the "Microsoft.Kiota.Serialization.Json" project (but not the generated files that cause the warning), and it restores the warnings that were switched off by ".globalconfig" (even for generated files).

src/serialization/json/.globalconfig
baywet: my question was more specifically: does it neutralize the warning for the whole JSON project? if so I think we should try to make it more specific to the instance of the issue to avoid potential regression when further changes are made to the project.

src/abstractions/serialization/IParseNode.cs
baywet: *Technically* IronRuby (last release in 2011) and IronPython (last release in 2024), as well as *really early* versions of VB.net... But I'm not sure we need to cater to those audiences.

And yes, to your point, callers would have to have a corresponding public property of the same type in their generated code, which would trigger the warning if their project is marked as CLSCompliant. Which means they'd be "aware of the issue" at the very least.

I don't have any objection for the inline suppression here. Thank you for the additional information.

src/serialization/json/.globalconfig
WolfgangHG: Yes, it neutralizes the warning for the whole JSON project. But the ".editorconfig" file is on the same level and restores it again for all ".cs" files (without the generated files).

baywet (approved):

PR #708, opened 2026-05-01
feat!: update target frameworks to net8.0 and net10.0, drop net5.0 and net6.0

MIchaelMainer:
# PR Summary

Update target frameworks to net8.0 and net10.0, drop net5.0 and net6.0. Updates target frameworks across the solution to align with currently supported .NET runtimes; net10.0 is the current LTS release. Target framework management for the seven common source libraries is consolidated into src/Directory.Build.props to prevent future drift.

Remove obsolete APIs and provided upgrade guide.

## Target Framework Changes (includes breaking change to target frameworks)

- **TFMs updated**: `netstandard2.0;netstandard2.1;net8.0;net10.0` (dropped `net5.0` and `net6.0`)
- **Test projects**: Target `net10.0` only
- **CI**: Updated SDK installs from .NET 6 → .NET 8 (for ESRP signing) + .NET 10, tests run on `net10.0`

### 1. `IAsyncParseNodeFactory` deleted

Merged into `IParseNodeFactory`, which now has a single async method:

```csharp
Task<IParseNode> GetRootParseNodeAsync(string contentType, Stream content, CancellationToken cancellationToken = default);
```

The old synchronous `GetRootParseNode()` method is removed entirely.

### 2. Synchronous deserialization methods removed

All synchronous `Deserialize`/`DeserializeCollection` methods removed from `KiotaSerializer` and `KiotaJsonSerializer` — only async variants remain.

### 3. `MultipartBody` overloads removed

`GetPartValue<T>(string)` and `RemovePart(string)` removed — the overloads requiring both `partName` and `contentType` are kept.

### 4. `ParseNodeFactoryRegistry` simplified

Now implements `IParseNodeFactory` (not `IAsyncParseNodeFactory`), and `GetFactory<T>()` simplified to non-generic `GetFactory(string contentType)`.

### 5. `ParseNodeProxyFactory` updated

Now implements `IParseNodeFactory` directly, with only the async `GetRootParseNodeAsync` method.

### 6. Serialization library factories updated

`JsonParseNodeFactory`, `FormParseNodeFactory`, `TextParseNodeFactory` now implement `IParseNodeFactory` instead of `IAsyncParseNodeFactory`, with only the async method.

## Build & Infrastructure

- **Central Package Management** added (Directory.Packages.props) consolidating all package versions
- **Package version lower bounds** updated to `8.0` for `System.Text.Json`, `System.Diagnostics.DiagnosticSource`, etc.
- **`Microsoft.CodeAnalysis.PublicApiAnalyzers`** restricted to run on `net8.0` only (avoids false positives from net10.0 source generator differences)
- **Test framework** migrated to xunit v3 with Microsoft Testing Platform
- **CLS compliance**: Added `#pragma warning disable CS3001/CS3002` around `sbyte` members in interfaces, and `[CLSCompliant]` attributes

## Trim/AOT Safety

- **IL2111 suppression** in `KiotaDeserializationWrapperFactory.Create` (method group loses `DynamicallyAccessedMembers` annotation)
- **IL2075 suppression** in `JsonSerializationWriter.WriteNonParsableObjectValue` (inherently trim-unsafe reflection path)

## Documentation

- New upgrade-guide-v1-to-v2.md with migration steps, compiler error mappings, and breaking change details


Directory.Packages.props
```diff
@@ -0,0 +1,26 @@
...
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+  <ItemGroup>
+    <!-- Source packages -->
+    <PackageVersion Include="Azure.Core" Version="1.50.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="5.3.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="5.3.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="4.14.0" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="[6.0,)" />
```
baywet: This range should also be updated

Directory.Packages.props
```diff
@@ -0,0 +1,26 @@
...
+    <PackageVersion Include="Microsoft.IO.RecyclableMemoryStream" Version="3.0.1" />
+    <PackageVersion Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.14.15" />
+    <PackageVersion Include="Std.UriTemplate" Version="2.0.8" />
+    <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="[8.0.1,)" />
+    <PackageVersion Include="System.Net.Http.WinHttpHandler" Version="[6.0,)" />
+    <PackageVersion Include="System.Text.Json" Version="[6.0.10,)" />
```
baywet: Same for this one

src/Directory.Build.props
```diff
@@ -30,7 +33,7 @@
       <PackagePath>
       </PackagePath>
     </None>
-    <PackageReference Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="4.14.0" Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)','net8.0'))">
+    <PackageReference Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Condition="'$(TargetFramework)' == 'net8.0'">
```
baywet, line 36: I believe the change in condition is a regression, won't enable for net10 for example

baywet (dismissed):
As we're making a new major revision, we should also take this opportunity to cleanup the interfaces. 
There is the async parse node interface to merge into the regular one and get rid of the sync method. 
There is also a cleanup to do with the static serialization methods and overloads for optional parameters. 
And generally check for additional to-dos

src/Directory.Build.props
gavinbarron: Nope, it's correct, there's a change in net10 where in the emitted code for this that the .g.cs file no longer has the #nullable enable directive at the top of the file thus creating an oblivious nullable context for the file.

MIchaelMainer:
I created another PR for:
* async parse node interface to merge into the regular one and get rid of the sync method.
* static serialization methods cleanup and overloads for optional parameters.
* to-dos
https://github.com/microsoft/kiota-dotnet/pull/709

src/Directory.Build.props
baywet: But this dependency is for the source generator for the version to work for telemetry I believe?

Directory.Packages.props
```diff
@@ -0,0 +1,26 @@
...
+    <PackageVersion Include="Azure.Core" Version="1.50.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="5.3.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="5.3.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="4.14.0" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="[8.0,)" />
+    <PackageVersion Include="Microsoft.IO.RecyclableMemoryStream" Version="3.0.1" />
+    <PackageVersion Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.14.15" />
+    <PackageVersion Include="Std.UriTemplate" Version="2.0.8" />
+    <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="[8.0.1,)" />
+    <PackageVersion Include="System.Net.Http.WinHttpHandler" Version="[6.0,)" />
```
baywet: I think the range here needs to be updated as well

MIchaelMainer:
@copilot resolve the merge conflicts in this pull request

MIchaelMainer:
> > @copilot resolve the merge conflicts in this pull request
> 
> Merge conflicts resolved in [c6176f0](https://github.com/microsoft/kiota-dotnet/commit/c6176f0b736ea48e0e32181265ce9312af4c2405). The two conflicting files were:
> 
> * `src/abstractions/Microsoft.Kiota.Abstractions.csproj` — added `CLSCompliantAttribute` from main, kept central package management (no explicit version on `Microsoft.VisualStudio.Threading.Analyzers`)
> * `src/http/httpClient/Microsoft.Kiota.Http.HttpClientLibrary.csproj` — added `CLSCompliantAttribute` from main, kept updated TFM conditions without `net5.0` and without old explicit version ranges

Same change as in https://github.com/microsoft/kiota-dotnet/issues/705

.azure-pipelines/ci-build.yml
```diff
@@ -13,15 +13,15 @@ pr:
...
+  buildPlatform: "Any CPU"
+  buildConfiguration: "Release"
 
 resources:
   repositories:
-  - repository: 1ESPipelineTemplates
-    type: git
-    name: 1ESPipelineTemplates/1ESPipelineTemplates
-    ref: refs/tags/release
+    - repository: 1ESPipelineTemplates
```
gavinbarron, line 21: the indentation changes in this file should be revisited

src/Directory.Build.props
gavinbarron: this is for the code analyzers that update/manage the shipped/unshipped API text listing AFAIK. 
Not sure how the telemetry and version interplay here, do you have any more insight?

.azure-pipelines/ci-build.yml
MIchaelMainer: This was caused by the VS Code formatter when we updated the dotnet version. The indented sequence style is also what we have in our other yaml files so this is providing yaml indent uniformity.

src/Directory.Build.props
MIchaelMainer: > I believe the change in condition is a regression, won't enable for net10 for example

The problem is that System.Text.Json's source generator (KiotaJsonSerializationContext) emits slightly different nullable annotations on net10.0 vs net8.0. Since PublicAPI.*.txt files are shared across all TFMs, the analyzer sees the net10.0-generated signatures as "different" from what's declared and raises false RS0016/RS0041 errors.

The options were:

* Maintain separate PublicAPI.*.txt files per TFM (complex, fragile)
* Suppress those specific warnings globally (hides real issues)
* Run the analyzer on only one TFM where the signatures match (**the chosen approach**)

Since the actual public API is identical across TFMs, running it on net8.0 alone catches all real public API changes. It's a pragmatic workaround for a source-generator quirk, not a fundamental incompatibility. 

We could update to net10 which would require us to update PublicAPI.*.txt files. Also, 8 is the lower common TFM supported.

MIchaelMainer:
@copilot I need a a commit message that describes all of the fixes, updates, and breaking changes in pull request 708. Use the guidance from https://github.com/googleapis/release-please#what-if-my-pr-contains-multiple-fixes-or-features

gavinbarron (approved):
:shipit:

Issue #481, opened 2024-11-21
Kiota C# Client crashes on Blazor WASM


brad-technologik:
### What are you generating using Kiota, clients or plugins?

API Client/SDK

### In what context or format are you using Kiota?

Nuget tool

### Client library/SDK language

Csharp

### Describe the bug

When I try and import my Kiota C# in a Blazor WASM Project, the page gets an error whenever the Kiota client is injected into a page:
```
ManagedError: One or more errors occurred. (Operation is not supported on this platform.)
```

### Expected behavior

Kiota C# client should work in Blazor. The error is kind of vague, but it may be related to multi threading which isn't supported in Blazor.

Given how Kiota and Blazor are both Microsoft products, I'd expect them work together.

### How to reproduce

```bash
dotnet new blazor -ai -int WebAssembly -n KiotaBlazorBug
cd KiotaBlazorBug/KiotaBlazorBug.Client
dotnet add package Microsoft.Kiota.Bundle
kiota generate --language=CSharp -d https://petstore3.swagger.io/api/v3/openapi.json -o ./Client -c PetStoreClient -n KiotaBlazorBug.Client
```
Add the following code to `Program.cs` (In *both* `KiotaBlazorBug/KiotaBlazorBug/Program.cs` and `KiotaBlazorBug/KiotaBlazorBug.Client`)

```cs
builder.Services.AddSingleton<PetStoreClient>(e =>
{
    var authProvider = new AnonymousAuthenticationProvider();
    // Create request adapter using the HttpClient-based implementation
    var adapter = new HttpClientRequestAdapter(authProvider);
    // Create the API client
    return new PetStoreClient(adapter);
});
```

Add the following line to `Counter.razor`

```cs
@inject PetStoreClient PetStoreClient
```

Run the project
```
cd ../KiotaBlazorBug
dotnet run
```

Visit http://localhost:5003/counter in your  Browser and observe the error and JS console.

### Open API description file

Any one. https://petstore3.swagger.io/api/v3/openapi.json is an example.

### Kiota Version

1.20.0+81bdb1bf13383a749f794d47926d09733d2d7be1

### Configuration

- OS: Mac OS 15.1 (24B83)

### Debug output

<details><summary>Click to expand log</summary>
```
blazor.web.js:1 crit: Microsoft.AspNetCore.Components.WebAssembly.Rendering.WebAssemblyRenderer[100]
      Unhandled exception rendering component: Operation is not supported on this platform.
[...]

PR #728, opened 2026-06-26
Added net10-browser as a target to enable the BROWSER directive

manuelelucchi:
Fixes a regression born from the upgrade to net10 as described [here](https://github.com/microsoft/kiota-dotnet/issues/481#issuecomment-4810069334)


baywet (approved):
Thanks for the contribution!

PR #805, opened 2026-09-17
Add support for the HTTP QUERY verb

ricardoboss:
Adds the QUERY method (RFC 10008) to the Method enum and allows streamed content for QUERY requests, which carry a body.

This relates to https://github.com/microsoft/kiota/issues/8126


baywet (approved):

baywet (commented):
Thanks, can we add tests for redirect and retry please?

baywet (approved):

